PR #57, opened 2016-11-29
Remove PLCrashReporter

achocron:
As it turns out, we don't need to include the PLCrashReporter framework binary at all; the only support we provide is with lc++ and iscxx flags. This means that the iOS Crashes NuGet package does not need to create the MobileCenterFrameworks folder at all. This PR prevents PLCrashReporter framework from being downloaded and updates the MobileCenterCrashes.a binary properties so it doesn't link with it.


Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/Contoso.Forms.Puppet.csproj
```diff
@@ -95,20 +95,6 @@
   <ItemGroup>
     <None Include="packages.config" />
   </ItemGroup>
-  <ItemGroup>
```
guperrot: Is that intentional?

PR #62, opened 2016-12-02
Feature/signal handler workaround

achocron:
This is necessary to fix an issue in iOS when certain signals (SIGBUS, SIGSEGV) cause an immediate crash when they shouldn't. When the crashes from these signals are legitimate, they will now not be captured by Mobile Center.


SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.iOS/CrashesExtension.cs
```diff
@@ -0,0 +1,16 @@
+﻿using Microsoft.Azure.Mobile.Crashes.iOS.Bindings;
+
+namespace Microsoft.Azure.Mobile.Crashes
+{   
+    public static partial class Crashes
```
guperrot: You don't need that class and reflection. Just move this call in static constructor of PlatformCrashes.

SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.Shared/Crashes.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.Azure.Mobile.Crashes
     /// <summary>
     /// Crashes service.
     /// </summary>
-    public static class Crashes
+    public static partial class Crashes
```
guperrot: Revert that, related to the comment on CrashesExtension

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.iOS/Contoso.Forms.Puppet.iOS.csproj
```diff
@@ -1,168 +1,168 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
guperrot: Do you need that change? Is that the current formatter rule or is it caused by a merge?

PR #63, opened 2016-12-02
Fix/multiple unrelated crashes

guperrot:


SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.Shared/Crashes.cs
```diff
@@ -36,6 +35,14 @@ static Crashes()
...
             PlatformCrashes.ShouldProcessErrorReport = null;
             PlatformCrashes.GetErrorAttachment = null;
+
+            /* 
+             * We need to add [Android.Runtime.Preserve] to BindingType to avoid it
+             * from being removed by "Link all assemblies optimization".
+             * However we cannot do it because this code is shared with ios and PCL.
+             * So instead we use the property explicitly here to preserve the method call even after optimization.
+             */
+            var type = BindingType;
```
achocron: Can you instead make `Crashes` a partial class and define `BindingType` in separate files for the two platforms (plus a PCL stub)? We are going to make `Crashes` a partial class in the signal handler workaround, anyway. That way we can use the `[Preserve]` attribute for `BindingType`  in `Crashes`, too.

guperrot: We should do a big refactoring to replace the PlatformCrashes stuff by partial. I would not mix both approaches together right now.

PR #68, opened 2016-12-03
Adjust wording of User Confirmation description

DanRigby:


README.md
```diff
@@ -210,7 +210,7 @@ Note that the events must be subscribed to and callbacks must be set before Mobi
 
         ```
 
-    * **User Confirmation:** If user privacy is important to you as a developer, you might want to get user confirmation before sending a crash report to Mobile Center. The SDK exposes a callback where you can tell it to await user confirmation before sending any crash reports.
+    * **User Confirmation:** By default the SDK automatically transmits crash reports to Mobile Center. However, the SDK exposes a callback where you can tell it to await user confirmation before sending any crash reports.
```
guperrot: We use "send" later to talk about callbacks, I'd suggest replacing transmits by sends.

DanRigby: 👍

PR #70, opened 2016-12-05
Don't crash in UWP

guperrot:
Display a message in the console instead.


guperrot:
There will be a second PR after this gets merged to add a demo app using internal repository for nuget (need a merge to develop for that).

PR #78, opened 2016-12-06
Add null-check to GetErrorAttachment callbacks

achocron:
We crash if the user implements GetErrorAttachment and returns null- this PR fixes that.

### DIFF
--- a/SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.iOS/CrashesDelegate.cs
+++ b/SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.iOS/CrashesDelegate.cs
@@ -31,7 +31,13 @@ namespace Microsoft.Azure.Mobile.Crashes
             }
               
             var report = ErrorReportCache.GetErrorReport(msReport);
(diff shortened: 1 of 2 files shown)


SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.iOS/CrashesDelegate.cs
```diff
@@ -31,7 +31,13 @@ public override MSErrorAttachment AttachmentWithCrashes(MSCrashes crashes, MSErr
             }
               
             var report = ErrorReportCache.GetErrorReport(msReport);
-            return _owner.GetErrorAttachment(report).internalAttachment;
+            ErrorAttachment attachment = _owner.GetErrorAttachment(report);
+            if (attachment != null)
```
guperrot: Use `attachment?.internalAttachment` next time ;)

achocron: I keep forgetting about that haha

PR #79, opened 2016-12-06
Add IntPtr constructor to CrashesInitializationDelegate

achocron:
This PR fixes an issue where starting an app with crashes disabled causes a crash.


SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.iOS.Bindings/Extensions.cs
```diff
@@ -18,6 +19,15 @@ private enum Signal
             SIGSEGV = 11
         }
 
+
+        public CrashesInitializationDelegate(IntPtr handle) : base(handle)
+        {
```
guperrot: Can you add comments why we need those constructors so that we don't delete them?

PR #92, opened 2016-12-08
Fix crash buttons on demo for release mode

guperrot:
And remove the one that test it does not crash.


guperrot:
Puppet will keep those advanced buttons as is.

PR #93, opened 2016-12-09
Expose IsConfigured in PCL

guperrot:


achocron:
Is there any reason why you didn't make it a property? Similar features like <service>.IsEnabled() are exposed as properties (<service>.Enabled) in the PCL.

PR #101, opened 2016-12-14
Fix ios crash when using "Link All" optimization

guperrot:


guperrot:
Need to update start-version script before merging.

PR #125, opened 2017-01-19
Hide error attachment functionality

achocron:


Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.Droid/Contoso.Forms.Puppet.Droid.csproj
```diff
@@ -27,6 +27,7 @@
     <WarningLevel>4</WarningLevel>
     <AndroidLinkMode>None</AndroidLinkMode>
     <AndroidSupportedAbis>arm64-v8a;armeabi;armeabi-v7a;x86;x86_64</AndroidSupportedAbis>
+<EmbedAssembliesIntoApk>false</EmbedAssembliesIntoApk>
```
guperrot: remove that change

SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes/ErrorAttachment.cs
```diff
@@ -3,19 +3,19 @@
...
     /// </summary>
-    public class ErrorAttachment
+    internal class ErrorAttachment
     {
         /// <summary>
         /// Gets or sets a plain text attachment.
         /// </summary>
         /// <value>The text attachment.</value>
-        public string TextAttachment { get; set; }
+        internal string TextAttachment { get; set; }
```
guperrot: that change should not needed if class itself is changed to internal

Tests/Contoso.Forms.Test.Strings/Strings.cs
```diff
@@ -58,7 +58,7 @@ public static class  TestStrings
         public static readonly string FailedToSendErrorReportLabel = "FailedToSendErrorReportLabel";
         public static readonly string ShouldProcessErrorReportLabel = "ShouldProcessErrorReportLabel";
         public static readonly string ShouldAwaitUserConfirmationLabel = "ShouldAwaitUserConfirmationLabel";
-        public static readonly string GetErrorAttachmentLabel = "GetErrorAttachmentLabel";
+        //public static readonly string GetErrorAttachmentLabel = "GetErrorAttachmentLabel";
```
guperrot: Should try to reduce to a minimum what we comment out, will be harder to restore the feature after otherwise.

guperrot:
Also need to remove the README documentation @achocron

PR #129, opened 2017-01-24
Update Android SDK to v0.5.0

guperrot:


SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes/ErrorBinaryAttachment.cs
```diff
@@ -9,18 +9,18 @@ internal class ErrorBinaryAttachment
         ///     Gets the binary data file name.
         /// </summary>
         /// <value>The name of the binary data file.</value>
-        internal string FileName { get; }
+        public string FileName { get; }
```
guperrot: That change was a comment I forgot to do on the previous PR: if the class is internal we didn't need to change visibility on fields, plus the change was not made in the other attachment class, this restores the consistency.

PR #139, opened 2017-02-17
Fix crash in Contoso.Forms.Puppet UWP

MatkovIvan:
Contoso.Forms.Puppet crashed in UWP when user open the Analytics tab and try to add a property.

I opened issue in Xamarin.Forms for crash when Navigation.PopModalAsync called on page with TableView.
For temp workaround for the crash, TableView replaced to StackLayout.


dhei:
Hi @MatkovIvan, do you have the link to the issue you opened on Xamarin.Forms?

MatkovIvan:
@dhei I opened the issue without sign in and do not have a link.

PR #141, opened 2017-02-24
Rename setServerUrl to setLogUrl in PCL

guperrot:


paulceli:
Hey,
What's the purpose of SetLogUrl?  it's not really documented..
I can't find a default value, so I guess it's not something that we could use to redirect traffic from clients through our reverse proxy to AppCenter.. or is it ?

thanks

guperrot:
@paulceli we use it internally to test against alternate endpoints like our dev and preprod ones and we didn't plan to document it. 

It could theoretically be used to proxy. The reason you don't find the default value in this repo is that our Xamarin SDK uses native bindings to our Android and Apple SDKs where the constant is defined there.

If you were looking to proxy, we don't provide support for this officially but you can look what the bound SDK uses for URL and payload at https://github.com/Microsoft/AppCenter-SDK-Android/blob/develop/sdk/appcenter/src/main/java/com/microsoft/appcenter/ingestion/AppCenterIngestion.java.

You can also increase log level to verbose using `AppCenter.LogLevel = Log.Verbose` before `AppCenter.start` and look at the Android logs to see all the https traces to debug.

PR #147, opened 2017-03-02
Move to iOS SDK 0.5.0

guperrot:
@achocron Don't merge label: by updating bindings we can't attach the .NET exception to the apple crash anymore. I have no idea why, needs investigation.


guperrot:
Now waiting for ios release to restart the build and test again locally.

PR #150, opened 2017-03-03
UWP Ingestion Unit Tests

MatkovIvan:
Added unit tests to Ingestion module. For isolate from network wrap system API to `HttpNetworkAdapter` and `NetworkStateAdapter`.

@achocron, this is only a preview, please don't merge yet. I plan add more test cases. But please let me know if there is anything which should be updated.


Tests/Microsoft.Azure.Mobile.Test.UWP/Ingestion/Http/NetworkStateTest.cs
```diff
@@ -0,0 +1,99 @@
...
+		}
+
+		private class TestNetworkStateAdapter : INetworkStateAdapter
+		{
+			public bool IsConnected { get; set; }
+
+			public event NetworkAddressChangedEventHandler NetworkAddressChanged;
+
+			public void TestNetworkAddressChanged()
+			{
```
achocron: Maybe there should be a test where ingestion tries to send logs when the network is offline, and then after that the network comes back online. (In that scenario, the logs should be sent when it comes back)

Tests/Microsoft.Azure.Mobile.Test.UWP/Ingestion/Http/IngestionHttpTest.cs
```diff
@@ -0,0 +1,121 @@
...
+		private IngestionHttp _ingestionHttp;
+
+		[TestInitialize]
+		public void InitializeIngestionHttpTest()
+		{
+			_adapter = Mock.Create<IHttpNetworkAdapter>();
+			_ingestionHttp = new IngestionHttp(_adapter.Object);
+		}
+		
+		[TestMethod]
```
achocron: Could you please add ///\<summary\> style comments to the test methods to briefly describe what they are trying to test?

MatkovIvan:
Addressed CR notes:
* Added more unit tests
* Encountered w/ the `HyberMock` issue which does not allow you to correctly calculate the num of method visits if you execute the same method with different input params => so I was unable to fully cover by unt tests `RetryableIngestion` class.

I propose accept this PR and then take a look on existing `HyberMock` limitations and see if we want to switch to another test framework.

achocron:
Could you add the unit test(s) for retry logic, even though there is a hyper mock bug? Then just mark with the [Ignore] attribute

PR #155, opened 2017-03-10
Implement telemetry logic

erychagov:
Without tests for now


erychagov:
Create a special branch for new feature

PR #156, opened 2017-03-13
Contoso.Forms.Puppet.UWP crashes fixed when building using native toolchain

bmourat:
The problem was that MissingMetadataException was thrown when app was build using native toolchain.
MissingMetadataExceptions happen in certain reflection scenarios because of new optimizations in the .NET Native compiler. More about it here: https://blogs.msdn.microsoft.com/dotnet/2014/05/21/net-native-deep-dive-help-i-hit-a-missingmetadataexception/.
Tha basic idea is that some stuff is thrown out when compiling, so we need to ensure everything stays using runtime directives in project .rd.xml file. But none of the solutions proposed in the article worked on our project and i assumed it was some wierd bug that appeared during creation or modification of project somehow. So one solution i've found was to create another project and replace original, and the bug dissapeared.

### DIFF
--- a/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.UWP/App.xaml.cs
+++ b/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.UWP/App.xaml.cs
@@ -1,18 +1,9 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.IO;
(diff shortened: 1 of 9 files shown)


achocron:
Works for me now, but when I compile in Debug and uncheck .NET native toolchain, the app doesn't seem to start.

PR #157, opened 2017-03-13
Added telemetry

erychagov:
The idea is to send the log MobileCenter.Start API call to our backend which includes the modules started by the SDK.

### DIFF
--- a/SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Ingestion/Models/StartServiceLog.cs
+++ b/SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Ingestion/Models/StartServiceLog.cs
@@ -6,6 +6,8 @@ namespace Microsoft.Azure.Mobile.Ingestion.Models
 
     public class StartServiceLog : Log
     {
+        internal const string JsonIdentifier = "StartServiceLog";
+
         /// <summary>
         /// Initializes a new instance of the Log class.
         /// </summary
@@ -26,7 +28,7 @@ namespace Microsoft.Azure.Mobile.Ingestion.Models
         /// feature).
         /// Concrete types like StartSessionLog or PageLog are always part of a
         /// session and always include this identifier.</param>
-        public StartServiceLog(long toffset, Mobile.Ingestion.Models.Device device, IEnumerable<string> services, System.Guid? sid = default(System.Guid?))
+        public StartServiceLog(long toffset, Device device, IEnumerable<string> services, System.Guid? sid = default(System.Guid?))
             : base(toffset, device, sid)
         {
             Services = new List<string>(services);
@@ -37,5 +39,21 @@ namespace Microsoft.Azure.Mobile.Ingestion.Models
         /// </summary>
         [JsonProperty(PropertyName = "services")]
         public List<string> Services { get; set; }
+
+        /// <summary>
+        /// Validate the object.
+        /// </summary>
+        /// <exception cref="Microsoft.Rest.ValidationException">
+        /// Thrown if validation fails
+        /// </exception>
+        public override void Validate()
+        {
+            base.Validate();
+
+            if (Services == null)
+            {
+                throw new Rest.ValidationException(Rest.ValidationRules.CannotBeNull, nameof(Services));
+            }
+        }
     }
 }
--- a/SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Ingestion/Models/StartServiceLog.cs
+++ b/SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Ingestion/Models/StartServiceLog.cs
@@ -4,6 +4,7 @@ namespace Microsoft.Azure.Mobile.Ingestion.Models
 {
     using Newtonsoft.Json;
 
+    [JsonObject( JsonIdentifier )]
     public class StartServiceLog : Log
     {
         internal const string JsonIdentifier = "StartServiceLog";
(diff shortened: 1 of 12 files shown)


SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/MobileCenter.cs
```diff
@@ -292,32 +296,42 @@ internal void StartInstance(params Type[] services)
...
+            if (startedServiceNames.Count > 0)
+            {
+                StartServiceLog serviceLog = new StartServiceLog();
+                serviceLog.Services = startedServiceNames;
+                _channelGroup.Enqueue( serviceLog );
+            }
         }
 
-        private void StartService(IMobileCenterService service)
+        private bool StartService(IMobileCenterService service)
```
achocron: Instead, could you keep this as void and change the service names container to a HashSet? That way, errors are still dealt with by using exceptions

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Ingestion/Models/StartServiceLog.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System.Collections.Generic;
+
```
achocron: The other concrete log classes have an attribute here for the json identifier - should it be here, too? It's used for type identification when deserializing derived types

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/MobileCenter.cs
erychagov: As I see in any way errors are still dealt with by using exceptions.
I have to add only service name which has not been started before. How I can know about it without return value?

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Ingestion/Models/StartServiceLog.cs
erychagov: Do you mean [this identifier](https://github.com/Microsoft/mobile-center-sdk-dotnet/pull/157/files/#diff-9dc6c835d2b9e5798e8d1e826c2d67d6R9)?

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/MobileCenter.cs
achocron: Could you add the names into a hash set instead of a list? That automatically doesn't have duplicates

achocron: Or you can still use a list, right? If an exception is thrown in start service, it won't be added because the exception will be thrown before the service can be added to the list. Instead of logging a message in the StartService method, just throw a MobileCenterException that contains that message

Tests/Microsoft.Azure.Mobile.Test.Windows/StartServiceLogTests.cs
```diff
@@ -0,0 +1,53 @@
+﻿using Microsoft.Azure.Mobile.Ingestion.Models;
+using Microsoft.Azure.Mobile.Utils;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.Mobile.Test.Windows
+{
+    [TestClass]
+    public class StartServiceLogTests
```
achocron: Rename to StartServiceLogTest

Tests/Microsoft.Azure.Mobile.Test.Windows/StartServiceLogTests.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        private const string StorageTestChannelName = "storageTestChannelName";
+
+        [TestInitialize]
+        public void asdf()
```
achocron: Rename to InitializeStartServiceTest

Tests/Microsoft.Azure.Mobile.Test.Windows/StartServiceLogTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            LogSerializer.AddFactory(StartServiceLog.JsonIdentifier, new LogFactory<StartServiceLog>());
+        }
+
+        [TestMethod]
```
achocron: Add comments to the [TestMethod]s that explain what they do (briefly, see other test files for example of this)

Tests/Microsoft.Azure.Mobile.Test.Windows/StartServiceLogTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            StartServiceLog log = new StartServiceLog();
+            Assert.IsNull(log.Device);
+            Assert.IsNull(log.Services);
+            Assert.IsNull(log.Sid);
+            Assert.AreEqual(0, log.Toffset);
+        }
+
+        [TestMethod]
+        public void SaveStartServiceLog()
```
achocron: Maybe this should be split into different test cases? It seems like it is testing a lot of different things in one method

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Ingestion/Models/StartServiceLog.cs
achocron: Yes but that's part of it- the class should also have the [JsonObject(JsonIdentifier)] attribute

erychagov:
@achocron please, review my changes.

PR #167, opened 2017-03-23
Improve code coverage

erychagov:


Tests/Microsoft.Azure.Mobile.Test.Windows/Channel/ChannelTest.cs
```diff
@@ -154,6 +155,102 @@ public void ChannelInvokesFailedToSendLogEvent()
...
+            _channel.SetEnabled(true);
+
+            Assert.IsTrue(SendingLogOccurred(MaxLogsPerBatch));
+        }
+
+        /// <summary>
+        /// Validate that FailedToSendLog calls when channel is disabled
+        /// </summary>
+        [TestMethod]
+        public void ChannelInvokesFailedToSendLogEventAfterEnabling()
```
achocron: Should the name of this be after disabling? Also, when I run this test on its own it passes, but when I select "run all" tests in visual studio, this one fails.

PR #173, opened 2017-03-31
Distribute iOS

TroubleMakerBen:


SDK/MobileCenterDistribute/Microsoft.Azure.Mobile.Distribute.iOS/Distribute.cs
```diff
@@ -10,14 +10,36 @@ public static partial class Distribute
...
         {
+            iOSDistribute.SetApiUrl(apiUrl);
+        }
+
+        /// <summary>
+        /// Process URL request for the Distribute service.
+        /// Place this method call into app delegate openUrl method.
+        /// </summary>
+        /// <param name="url">The url with parameters.</param>
+        public static void PlatformOpenUrl(NSUrl url)
```
guperrot: Platform prefix for method name is only used because of the partial class pattern with portable. Since this is a platform specific API and public and documented, remove Platform prefix from this one.

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.iOS/AppDelegate.cs
```diff
@@ -15,6 +16,13 @@ public override bool FinishedLaunching(UIApplication uiApplication, NSDictionary
             LoadApplication(new App());
             return base.FinishedLaunching(uiApplication, launchOptions);
         }
+
+        public override bool OpenUrl(UIApplication application, NSUrl url, string sourceApplication, NSObject annotation)
+        {
+            MSDistribute.OpenUrl(url);
```
guperrot: We don't document bindings, please use only the wrapper c# API we add in test apps.

TroubleMakerBen:
@guperrot addressed your comments

guperrot:
Will approve after testing the real thing :)

PR #177, opened 2017-04-05
Fix comment

guperrot:


guperrot:
@achocron Yeah just needed a reason to push something to trigger vsts ;)

PR #178, opened 2017-04-05
Feature/update readme

TroubleMakerBen:


TroubleMakerBen:
This should be ready to be merged!

README.md
```diff
@@ -95,37 +98,78 @@ To start the SDK in your app, follow these steps:
...
     ```csharp
-    MobileCenter.Start("ios={Your Xamarin iOS App Secret};android={Your Xamarin Android App secret}",typeof(Analytics), typeof(Crashes));
+    MobileCenter.Start("ios={Your Xamarin iOS App Secret};android={Your Xamarin Android App secret}",typeof(Analytics), typeof(Crashes), typeof(Distribute));
     ```
      
     You need to copy paste the App secret value for Xamarin iOS and Android app from Mobile Center portal. Make sure to replace the placeholder text above with the actual values for your application.
     
     The example above shows how to use the `Start()` method and include both the Analytics and Crashes module. If you wish not to use Analytics, remove the parameter from the method call above. Note that, unless you explicitly specify each module as parameters in the start method, you can't use that Mobile Center service. Also, the `Start()` API can be used only once in the lifecycle of your app – all other calls will log a warning to the console and only the modules included in the first call will be available.
+    
+   **Note that your forms iOS app requires additional setup steps for in-app updates to work.**
```
guperrot: Replace `forms iOS` by `Xamarin.iOS`

README.md
```diff
@@ -251,8 +295,36 @@ Note that the events must be subscribed to and callbacks must be set before Mobi
...
+
+### Localization of the update UI
+
+You can easily provide your own resource strings if you'd like to localize the text displayed in the update dialog. Look at the string files for iOS [in this resource file](https://github.com/Microsoft/mobile-center-sdk-ios/blob/develop/MobileCenterDistribute/MobileCenterDistribute/Resources/en.lproj/MobileCenterDistribute.strings) and those for Android [in this resource file](https://github.com/Microsoft/mobile-center-sdk-android/blob/distribute/sdk/mobile-center-distribute/src/main/res/values/strings.xml).
+
+Use the same string name and specify the localized value to be reflected in the dialog in your own app resource files.  
+
+### Enable or disable Distribute
+
+You can change the enabled state by calling the `setEnabled` API. If you disable it, the SDK will not prompt your users when a new version is available for install. To re-enable it, pass `YES` or `true` as a parameter in the same method.
```
guperrot: setEnabled => its not a method on c# its a property as in  `Mobile.Enabled = false`

guperrot: And there is no YES

README.md
```diff
@@ -251,8 +295,36 @@ Note that the events must be subscribed to and callbacks must be set before Mobi
         };
 
         ```
+## 7. Distribute APIs
+
+You can easily let your users get the latest version of your app by integrating Mobile Center Distribute. All you need to do is pass the service name as a parameter in the `start()` API call. Once the app launches, the SDK checks for new updates in the background. If it finds a new update, users will see a dialog with three options - **Download**, **Postpone** and **Ignore**. If the user taps **Download**, Mobile Center Distribute will trigger the new version to be installed. **Postpone** will delay the download until the app is opened again. **Ignore** will not prompt the user again for that particular app version. In case of a required update, the user will only have the option to tap **Download**.
```
guperrot: start => Start

TroubleMakerBen:
Addressed comments

PR #185, opened 2017-04-10
Consolidate NETCore UniversalWindowsPlatform reference to 5.2.2

dhei:
This change is to make it compatible for Visual Studio 2015


guperrot:
Restarted VSTS build manually and it passed.

PR #186, opened 2017-04-11
Add a script to trigger PR VSTS build from bitrise

guperrot:


guperrot:
@achocron to avoid a git clone on bitrise I inlined the script there, just saving it in git not to lose it.

guperrot:
This PR was run by the bitrise to VSTS script

PR #191, opened 2017-04-12
Demo pre release

guperrot:


dhei:
I found the same issue as #139 in Contoso.Forms.Demo app. I pushed one commit to fix the TableView problem Navigation.PopModalAsync in Xamarin.Forms.

PR #208, opened 2017-04-17
Custom properties feature

MatkovIvan:
Implemented custom properties feature by spec

Same as https://github.com/Microsoft/mobile-center-sdk-android/pull/387 and https://github.com/Microsoft/mobile-center-sdk-ios/pull/449


guperrot:
Looks good apart for the double signature that would really be nice to have as mentioned above.
Will approve after that.

PR #214, opened 2017-04-19
Intellisense UWP

achocron:
We had an issue where if you created a new UWP app and added our packages it would build, but VS would think there are missing assemblies. Also removed target filenames from nuspecs as they are not needed


guperrot:
@dhei needed to test xamarin before merge, I hope it won't break ;)

PR #230, opened 2017-04-25
Convert Forms Demo to project.json

MatkovIvan:


Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.iOS/project.json
```diff
@@ -0,0 +1,15 @@
+{
+  "dependencies": {
+    "Microsoft.Azure.Mobile": "0.9.0",
```
MatkovIvan: This is for iOS

guperrot: That comment was to harmonize with all platforms since explicit core dependency is not needed.

guperrot:
I will check if we can build in Mobile Center with project.json before merge.

PR #292, opened 2017-05-16
Merge branch 'develop' into winforms

MatkovIvan:


SDK/MobileCenter/Microsoft.Azure.Mobile.WinForms/Utils/ApplicationLifecycleHelper.cs
```diff
@@ -104,8 +111,13 @@ public bool Enabled
             }
         }
 
+        public bool HasShownWindow => started && !suspended;
```
achocron: Should just map to started because the property only indicates whether UI has been shown at some point

SDK/MobileCenter/Microsoft.Azure.Mobile.WinForms/Utils/ApplicationLifecycleHelper.cs
```diff
@@ -34,6 +36,11 @@ private static void WinEventHook(IntPtr winEventHookHandle, uint eventType, IntP
                 return;
             }
             var anyNotMinimized = Application.OpenForms.Cast<Form>().Any(form => form.WindowState != FormWindowState.Minimized);
+            if (!started && anyNotMinimized)
+            {
+                started = true;
+                Start?.Invoke();
+            }
             if (suspended && anyNotMinimized)
```
achocron: Should be `else if` so that we don't invoke the resuming event at startup. Should maybe also set `suspended = false` in the above conditional block?

SDK/MobileCenter/Microsoft.Azure.Mobile.WinForms/Utils/ApplicationLifecycleHelper.cs
achocron: Should only be mapped to `started` because the property is to indicate whether UI has been shown at any point in the past

SDK/MobileCenter/Microsoft.Azure.Mobile.WinForms/Utils/ApplicationLifecycleHelper.cs
MatkovIvan: Generally speaking nothing will be changed. We don't call resume event on start now.

PR #293, opened 2017-05-16
Feature/uwp unit test projects

achocron:


achocron:
Had to close and reopen - I think VSTS failed because NuGet version was too old to restore packages with project.json. Updated it and it seems to work now.

PR #294, opened 2017-05-17
Uwp push callback

guperrot:


guperrot:
Should be merged only after backend deployed to prod.

SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
```diff
@@ -1,67 +1,174 @@
...
+        /// </summary>
+        private void ApplyEnabledState()
+        {
+            // Since the lock we use is not recursive, caller of this method is expected to execute this method inside lock
+            if (Enabled)
             {
-                MobileCenterLog.Warn(MobileCenterLog.LogTag, "Push service registering with Mobile Center backend has failed", e);
-                return;
+                var stateSnapshot = _stateKeeper.GetStateSnapshot();
+                Task.Factory.StartNew(async () =>
```
achocron: I think moving the body of this lambda into its own method would make this much more readable, but I won't press the issue either way

achocron:
Could you add the project "Microsoft.Azure.Mobile.Push.Shared" to the Windows-Build solution? It isn't required, but since it's referenced by one of the projects in that solution, I think it should be added. (Even though doesn't directly pertain to your changes)

SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
```diff
@@ -1,67 +1,174 @@
...
+        private void OnPushNotificationReceivedHandler(PushNotificationChannel sender, WindowsPushNotificationReceivedEventArgs e)
+        {
+            if (e.NotificationType == PushNotificationType.Toast)
             {
-                MobileCenterLog.Debug(MobileCenterLog.LogTag, $"Push token '{pushToken}'");
-
-                var pushInstallationLog = new PushInstallationLog(0, null, pushToken, Guid.NewGuid());
-
-                Channel.Enqueue(pushInstallationLog);
+                var content = e.ToastNotification.Content;
```
achocron: Should any part of this be null checked?

SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
```diff
@@ -1,67 +1,174 @@
...
             {
-                MobileCenterLog.Error(LogTag, "Push service registering with Mobile Center backend has failed.");
+                MobileCenterLog.Debug(LogTag, $"Push ignored. We only handle Toast notifications but PushNotificationType is '{e.NotificationType}'.");
             }
-
-            _mutex.Unlock();
         }
 
-        private async Task<PushNotificationChannel> CreatePushNotificationChannel()
+        private PushNotificationReceivedEventArgs ParseMobileCenterPush(XmlDocument content)
```
achocron: Should this be made static?

SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
```diff
@@ -1,67 +1,174 @@
...
+                            customData.Add(pair.Key, pair.Value.ToString());
+                        }
+                        return customData;
+                    }
+                }
+                return null;
+            }
+            catch (JsonReaderException)
+            {
+                return null;
```
achocron: In general, would it not be better to ignore these low level exceptions and have them caught at the highest level possible so that we can eliminate many of the null checks and delegate the error control flow logic to the built in exception mechanisms? Not sure what the right answer is

SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
guperrot: I don't think so.

SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
guperrot: If it can it should yes ;)

SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
guperrot: This one is a particular case as we expect to have launch strings from customer code that are not mobile center. The only way for us to check it's for us is parsing as json.

PR #301, opened 2017-05-17
Fix locking in Analytics and Push for UWP

guperrot:


guperrot:
Don't merge as there is a bug in UI not showing in forms puppet...

PR #317, opened 2017-05-22
Make internal APIs that do work on background threads return tasks

MatkovIvan:
- `Channel.SetEnabled` -> `Channel.SetEnabledAsync`;
- `Channel.Shutdown` -> `Channel.ShutdownAsync`;
- `Channel.Enqueue` -> `Channel.EnqueueAsync`;
- Ignore `StatefulMutexException` in `Channel.CountFromDiskAsync`;
- Call `CountLogs` on channel enable;
- `ServiceCall.Execute` -> `ServiceCall.ExecuteAsync`;
- Use async task instead `ServiceCallSucceededCallback`/ `ServiceCallFailedCallback ` in `IServiceCall`. This allows simplify tests;


SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -104,37 +109,30 @@ public bool IsEnabled
...
                     _mutex.Unlock();
                     SendingLog?.Invoke(this, new SendingLogEventArgs(log));
                     FailedToSendLog?.Invoke(this, new FailedToSendLogEventArgs(log, new CancellationException()));
-                    return null;
                 }
                 _mutex.Unlock();
                 EnqueuingLog?.Invoke(this, new EnqueuingLogEventArgs(log));
-                _mutex.Lock(stateSnapshot);
-                return Task.Run(async () =>
+                await Task.Run(async () =>
```
achocron: Now that enqueue is async, we don't need to wrap these in a Task.Run call

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -104,37 +109,30 @@ public bool IsEnabled
...
                 {
                     await PrepareLogAsync(log).ConfigureAwait(false);
```
achocron: This probably should have been done before, but PrepareLogAsync should take a statesnapshot parameter, and lock around the setting of _device. But the call to GetDeviceInformationAsync should be done before the lock is acquired

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -317,49 +307,37 @@ private Task Suspend(bool deleteLogs, Exception exception)
...
-                _mutex.Unlock();
+                _sendingBatches.Add(batchId, logs);
+                _pendingLogCount -= logs.Count;
+                await TriggerIngestion(logs, stateSnapshot, batchId).ConfigureAwait(false);
+                await CheckPendingLogs(stateSnapshot).ConfigureAwait(false);
             }
         }
 
-        private void TriggerIngestion(IList<Log> logs, State stateSnapshot, string batchId)
+        private async Task TriggerIngestion(IList<Log> logs, State stateSnapshot, string batchId)
```
achocron: TriggerIngestionAsync

achocron:
If the method is async, please change the name to end with Async

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
MatkovIvan: `TriggerIngestionAsync` already exists. I'll rename it to another.

achocron:
In Channel (and possibly other areas, but still reviewing): Whenever a call to await is made, please release the lock and then reacquire it after the call (with the statesnapshot parameter) - this way calls to Lock() will not block because we won't hold it for long periods of time

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -436,53 +401,61 @@ private void HandleSendingFailure(string batchId, IngestionException e)
                     FailedToSendLog?.Invoke(this, new FailedToSendLogEventArgs(log, e));
                 }
             }
-            _mutex.Lock();
-            try
+            Suspend(!isRecoverable, e);
```
achocron: await Suspend

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -190,47 +188,50 @@ public void InvalidateDeviceCache()
...
+                // We will call CountLogs when channel will be enabled again.
             }
             finally
             {
                 _mutex.Unlock();
             }
         }
 
-        private Task Suspend(bool deleteLogs, Exception exception)
+        private async Task Suspend(bool deleteLogs, Exception exception)
```
achocron: SuspendAsync

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -46,9 +46,9 @@ public sealed class Channel : IChannelUnit
             Task.Run(CountFromDiskAsync);
```
achocron: If this is done in setenabled do we need to do it here?

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -368,62 +327,62 @@ private void TriggerIngestion(IList<Log> logs, State stateSnapshot, string batch
                 {
                     _mutex.Unlock();
                     SendingLog?.Invoke(this, eventArgs);
-                    _mutex.Lock(stateSnapshot);
+                    await CheckPendingLogsAsync(stateSnapshot).ConfigureAwait(false);
```
achocron: I guess this should be lockasync?

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -43,12 +43,11 @@ public sealed class Channel : IChannelUnit
             _batchScheduled = false;
             _enabled = true;
             DeviceInformationHelper.InformationInvalidated += (sender, e) => InvalidateDeviceCache();
-            Task.Run(CountFromDiskAsync);
```
achocron: I know my comment last time asked if we could delete this, but looking now I guess we actually do need it? Problem is that we don't necessarily call SetEnabledAsync(true) before using, so either we should (but can't in constructor) or we should restore this

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -177,151 +164,126 @@ public Task Enqueue(Log log)
...
+            catch (StatefulMutexException e)
+            {
+                MobileCenterLog.Warn(MobileCenterLog.LogTag, "The Suspend operation has been cancelled", e);
+            }
+            if (deleteLogs && FailedToSendLog != null)
+            {
+                foreach (var log in _sendingBatches.Values.SelectMany(batch => batch))
                 {
-                    _pendingLogCount = 0;
-                    return Task.Run(DeleteLogsOnSuspendedAsync);
```
achocron: Now that this line is gone, `DeleteLogsOnSuspendedAsync` must be called. Consider adding this after the `if (deleteLogs && FailedToSendLog != null)` block:
```
            else if (deleteLogs)
            {
                await DeleteLogsOnSuspendedAsync();
            }
```

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -336,158 +298,132 @@ private Task Suspend(bool deleteLogs, Exception exception)
...
             else if (_pendingLogCount > 0 && !_batchScheduled)
             {
                 _batchScheduled = true;
-                Task.Delay((int)_batchTimeInterval.TotalMilliseconds).ContinueWith(async completedTask =>
+
+                // No need wait _batchTimeInterval here.
+#pragma warning disable CS4014
+                Task.Run(async () =>
                 {
+                    await Task.Delay((int)_batchTimeInterval.TotalMilliseconds).ConfigureAwait(false);
```
achocron: Instead of wrapping this in a `Task.Run`, why not use `ContinueWith`? You don't need to `await` the result

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Utils/Synchronization/StatefulMutex.cs
```diff
@@ -56,31 +114,45 @@ public void Lock(State stateSnapshot)
...
+            _mutex.Release();
+        }
+
+        public LockHolder GetLock()
+        {
+            Lock();
+            return new LockHolder(this);
+        }
+
+        public async Task<LockHolder> GetLockAsync()
```
achocron: Why not just make Lock() return a LockHolder (and same for async versions)? That way we don't have more methods than needed

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Utils/Synchronization/StatefulMutex.cs
```diff
@@ -56,31 +114,45 @@ public void Lock(State stateSnapshot)
...
-        /// Asynchronously locks the mutex and does not verify any state
-        /// </summary>
-        /// <seealso cref="Lock"/>
-        public Task LockAsync()
-        {
-            return _mutex.WaitAsync();
-        }
-
-        /// <summary>
         /// Unlocks the mutex. Consecutive calls will not throw an exception
```
achocron: This summary is now invalid

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Utils/Synchronization/StatefulMutex.cs
```diff
@@ -11,16 +11,53 @@ namespace Microsoft.Azure.Mobile.Utils.Synchronization
...
+            get { return _state; }
+        }
 
         /// <summary>
-        /// Creates a <see cref="StatefulMutex"/>  with an existing <see cref="StateKeeper"/> 
+        /// Advances the current state
         /// </summary>
-        /// <param name="stateKeeper">Tracks the current state for the mutex</param>
-        public StatefulMutex(StateKeeper stateKeeper)
+        public void InvalidateState()
```
achocron: Maybe make this return the new state, and then we can change 
```
_mutex.InvalidateState();
state = _mutex.State;
```
to
```
state = _mutex.InvalidateState();
```

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
achocron: I'm confused because we enter channel in an enabled state and start using it without ever calling `ResumeAsync`. Should this be invoked when it is created? And if so, how will you ensure that it is called first? Perhaps create the mutex in a locked state, and then unlock it at the end of an initialization function (that doesn't lock it but assumes mutex is already locked)

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Utils/Synchronization/StatefulMutex.cs
```diff
@@ -11,16 +11,53 @@ namespace Microsoft.Azure.Mobile.Utils.Synchronization
...
+        /// Checks if the given state is current
+        /// </summary>
+        /// <param name="state">The state to test</param>
+        /// <returns>True if the current state is current, false otherwise</returns>
+        public bool IsCurrent(State state)
+        {
+            return _state.Equals(state);
+        }
+
+        public async Task ShutdownAsync()
```
achocron: I like the idea here, but I'm not convinced that it's correct. What if I have this:
Task A calls _mutex.Lock();
Task B calls await _mutex.Lock();
Task C calls await _mutex.Shutdown();

If Task C gets scheduled during task A to run next, then task B will never run. I'm not sure if this is the correct behavior. @guperrot can you comment on this?

Problem is that there is no guarantee that the tasks will be executed in order of when the call is made.

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Utils/Synchronization/StatefulMutex.cs
MatkovIvan: Good point

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Utils/Synchronization/StatefulMutex.cs
MatkovIvan: @achocron disagree.
`await _mutex.WaitAsync()` not equal `await Task,Run(() => _mutex.Wait())`

achocron: @MatkovIvan true, but isn't there still the problem of the semaphore being transferred out of order?

achocron: https://msdn.microsoft.com/en-us/library/system.threading.semaphoreslim(v=vs.110).aspx 

"If multiple threads are blocked, there is no guaranteed order, such as FIFO or LIFO, that controls when threads enter the semaphore. "

achocron: That's why the shutdown logic was so complicated before in storage

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Utils/Synchronization/StatefulMutex.cs
MatkovIvan: Nope. Just git diff strange sometimes

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
achocron: Now that I think about it I'm not sure

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Channel/Channel.cs
```diff
@@ -43,34 +41,12 @@ public sealed class Channel : IChannelUnit
...
-                else
-                {
-                    Suspend(true, new CancellationException());
-                }
-            }
-            finally
-            {
-                _mutex.Unlock();
-            }
+                _pendingLogCount = task.Result;
```
achocron: Please wrap this in a check to see if the task completed successfully

PR #321, opened 2017-05-25
Feature/shutdown logic

achocron:
Implemented a task queue in storage to simplify certain aspects of it


SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Storage/Storage.cs
```diff
@@ -224,7 +291,19 @@ await _storageAdapter.GetAsync<LogEntry>(entry => entry.Channel == channelName,
...
+            try
+            {
+                _queue.Add(task);
             }
+            catch (InvalidOperationException)
+            {
+                throw new StorageException("The operation has been cancelled");
+            }
+
+            await FlushQueue().ConfigureAwait(false);
```
guperrot: So basically if you keep putting logs, the queue will never reach 0 and get will never return ever.

guperrot: It would work better to just have a unique FlushQueue call that continously run in background and never return and remove all `await FlushQueue().ConfigureAwait(false);` calls to avoid that. Plus all of your `FlushQueue()` are competing with each other, this is overkill.

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Storage/Storage.cs
```diff
@@ -149,10 +185,27 @@ await _storageAdapter.DeleteAsync<LogEntry>(entry => entry.Channel == channelNam
...
-                return await _storageAdapter.CountAsync<LogEntry>(entry => entry.Channel == channelName)
-                    .ConfigureAwait(false);
+                throw new StorageException("The operation has been cancelled");
+            }
+            try
+            {
+                await FlushQueue().ConfigureAwait(false);
+                return await task.ConfigureAwait(false);
+            }
+            catch (AggregateException e)
```
dhei: >catch (AggregateException e) [](start = 12, length = 28)

Two questions: (1) which line throws AggregateException? under what condition? (2) Why AggregateException there's no try-catch for other places where FlushQueue() is called in this file?

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Storage/Storage.cs
```diff
@@ -149,10 +185,27 @@ await _storageAdapter.DeleteAsync<LogEntry>(entry => entry.Channel == channelNam
...
+            {
+                throw e.InnerException ?? new StorageException();
```
dhei: >throw e.InnerException ?? new StorageException(); [](start = 16, length = 49)

If you can add an error message to new StorageException(), should be helpful.

dhei:
Please ignore my previous comments. I found out I commented on a out-of-date version of your PR.

PR #322, opened 2017-05-26
Feature/combine with wpf

achocron:
Don't merge yet because I need to test the NuGet packages (suffix 0.12.1-r0004-20b4217). Also, still need to make SDK name detect current platform ("windowsclassic" is too general).

- adds a WPF puppet app
- changes WinForms SDK to be called WindowsClassic
- adds features in the WinForms SDK to make compatible with WPF (see lifecycle helper)


Apps/Contoso.WPF.Puppet/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,55 @@
...
+//
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
```
guperrot: Here we go again with this :)

Apps/Contoso.WinForms.Puppet/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
...
+//      Major Version
+//      Minor Version
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("0.0.0.0")]
+[assembly: AssemblyFileVersion("0.11.2.0")]
```
guperrot: Should run start-new-version 0.13.0 as WPF is a new feature, this will also fix most of your new assemblies I guess.

SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/Utils/AbstractDeviceInformationHelper.cs
```diff
@@ -74,13 +74,14 @@ protected virtual string GetLiveUpdatePackageHash()
 
         private string GetSdkVersion()
         {
-           return GetType().GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
+           return GetType().GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion;
```
guperrot: I am very concerned here. This field is mandatory. If we can't always have it we need to use a constant instead like in Xamarin.

Apps/Contoso.WPF.Puppet/Properties/AssemblyInfo.cs
achocron: gah I missed some

PR #325, opened 2017-05-26
Feature/xamarin custom properties

guperrot:


Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/App.xaml.cs
```diff
@@ -74,6 +74,12 @@ protected override void OnStart()
             MobileCenter.Start($"uwp={uwpKey};android={androidKey};ios={iosKey}",
                                typeof(Analytics), typeof(Crashes), typeof(Distribute), typeof(Push));
 
+            CustomProperties properties = new CustomProperties();
+            properties.Set("a", "b").Clear("a").Set("decimal", 3.000001M);
```
dhei: Can you add invalid cases in the test? For instance, key not matching regex, empty/null key, etc.

dhei:
Do you plan to add unit tests now or maybe later?

guperrot:
UWP has unit tests, xamarin is just bindings to unit tested SDKs, not sure about xamarin unit tests but it's out of scope of this PR.

@dhei please approve.

PR #332, opened 2017-06-01
Print warning to log if there more than 2 attachments per crash report

erychagov:


SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.Android/AndroidCrashListener.cs
```diff
@@ -7,6 +7,11 @@ namespace Microsoft.Azure.Mobile.Crashes
 {
     class AndroidCrashListener : Object, ICrashesListener
     {
+        /// <summary>
+        /// The max attachments per crash report.
+        /// </summary>
+        private const int kMaxAttachmentsPerCrashReport = 2;
```
guperrot: Please remove `k` from constants, we use PascalCase as suggested by Resharper.

SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.iOS/CrashesDelegate.cs
```diff
@@ -5,6 +5,11 @@ namespace Microsoft.Azure.Mobile.Crashes
 {
     public class CrashesDelegate : MSCrashesDelegate
     {
+        /// <summary>
+        /// The max attachments per crash report.
+        /// </summary>
+        private const int kMaxAttachmentsPerCrashReport = 2;
```
guperrot: Please remove `k` from constants, we use PascalCase as suggested by Resharper.

guperrot:
Oh wait, we don't need to it in Xamarin after bindings are updated. Sorry for that one, the check will be done by Android and iOS SDKs, this would duplicate the log.

PR #333, opened 2017-06-01
Hide MobileCenterLog from intellisense

MatkovIvan:
I don't used `Obsolete` attribute to avoid lot of warnings inside SDK. Globally suppress obsolete warning is a bad idea.

### DIFF
--- a/SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/MobileCenterLog.cs
+++ b/SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/MobileCenterLog.cs
@@ -1,4 +1,5 @@
 ﻿#define DEBUG
+using System.ComponentModel;
 using System.Diagnostics;
 
 namespace Microsoft.Azure.Mobile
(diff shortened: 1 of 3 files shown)


SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/MobileCenterLog.cs
```diff
@@ -8,13 +9,15 @@ public static partial class MobileCenterLog
...
         /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public static string LogTag { get; private set; }
         private static readonly object LogLock = new object();
         private static LogLevel _level = LogLevel.Assert;
 
         /// <summary>
         /// Gets or sets the log verbosity level.
         /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
guperrot: This one is internal, should remove attribute?

achocron:
@guperrot since the obsolete attribute wasn't added, people with ReSharper might still see the method (because ReSharper ignores the attribute by default). What are your thoughts here? We do use Obsolete on the other methods we hide

PR #334, opened 2017-06-02
Fix/synchronization issues in channel

achocron:
Some pretty nasty bugs addressed here. Couldn't repro issue with `KeyNotFoundException` but I could get something close to it at one point, which I've fixed here. Regardless, this should fix other issues that are tough to reproduce.


guperrot:
@achocron will review after conflicts solved

PR #345, opened 2017-06-07
Fix UI test login

guperrot:


scripts/mobile-center-login.sh
```diff
@@ -27,7 +24,7 @@ fi
 
 # Log in to Mobile Center
 echo "Logging in to mobile center..."
-mobile-center login -u "$USERNAME" -p "$PASSWORD"
+mobile-center login --token "$MOBILE_CENTER_API_TOKEN"
```
dhei: I have a naive question, should it be "-token" or "--token" in the command?

guperrot: Nope, usually command line params are either -t or --token, not -word (on unix -word woud generally mean -w -o -r -d)

dhei: Aha, thanks!

PR #350, opened 2017-06-14
Avoid duplicate start session when resuming with a log

guperrot:
Same as https://github.com/Microsoft/mobile-center-sdk-android/pull/456


Tests/Microsoft.Azure.Mobile.Analytics.Test.Windows/SessionTrackerTest.cs
```diff
@@ -55,6 +55,20 @@ public void ResumeAfterTimeout()
             _mockChannel.Verify(channel => channel.EnqueueAsync(It.IsAny<StartSessionLog>()), Times.Exactly(2));
         }
 
+        [TestMethod]
```
achocron: Could you please add a description for this test like we have for the others? Makes it easier when we need to modify them or debug failures.

PR #351, opened 2017-06-16
Feature/async getters

guperrot:


guperrot:
https://www.bitrise.io/build/811665d1eee6763d

achocron:
@guperrot ReSharper complains about everywhere there is a private member without explicitly marking it `private` because it wants us to be consistent. I know VS Mac/Xamarin Studio does the opposite. Can we go with the ReSharper rule?

guperrot:
@achocron I don't have resharper though. Maybe we could use private only on windows projects?

PR #359, opened 2017-06-23
Feature/Remove dependencies and use MSBuild

achocron:
Didn't end up fixing the .NET Standard bug, but a few other useful things came out of the investigation. Needs more thorough testing before merge


build.cake
```diff
@@ -370,7 +370,7 @@ Task("NuGet")
 		NuGetPack(spec, new NuGetPackSettings {
 			BasePath = basePath,
 			Verbosity = NuGetVerbosity.Detailed,
-			Version = module.NuGetVersion
+			Version = "0.5.0-test"//module.NuGetVersion
```
guperrot: what is that

nuget/MobileCenter.nuspec
```diff
@@ -21,6 +21,11 @@
         <dependency id="Newtonsoft.Json" version="6.0.1"/>
         <dependency id="sqlite-net-pcl" version="1.3.1"/>
       </group>
+      <!-- Use dependency groups for iOS and Android so that NETStandard dependencies aren't added to them -->
```
guperrot: update comment so that it's not .net standard specific and so we can merge even with the rollback

nuget/MacMobileCenter.nuspec
```diff
@@ -16,6 +16,11 @@
         <group targetFramework="netstandard1.0">
             <dependency id="NETStandard.Library" version="1.6.1" />
         </group>
+        <!-- Use dependency groups for iOS and Android so that NETStandard dependencies aren't added to them -->
```
guperrot: update comment so that it's not .net standard specific and so we can merge even with the rollback

.gitignore
```diff
@@ -33,6 +33,7 @@ bld/
 [Bb]in/
 [Oo]bj/
 Reference/
+AssemblyTFMAttribute.cs
```
guperrot: What is that?

PR #368, opened 2017-06-29
Fix null pointer exception when setting custom properties before Mobile Center is configured

bmourat:


SDK/MobileCenter/Microsoft.Azure.Mobile.Windows.Shared/MobileCenter.cs
```diff
@@ -255,6 +255,11 @@ private void SetInstanceLogUrl(string logUrl)
 
         private void SetInstanceCustomProperties(CustomProperties customProperties)
         {
+            if (!Configured)
+            {
+                MobileCenterLog.Error(MobileCenterLog.LogTag, "Mobile Center must be enabled before setting custom properties.");
```
guperrot: On Android it's
> Mobile Center hasn't been configured. You need to call MobileCenter.start with appSecret or MobileCenter.configure first.

(Also it's not about enabled but configured here).

PR #369, opened 2017-06-29
Fix ui thread blocker

achocron:
Fixes a severe block on UI thread where we wait for a lock on UI thread that is held on to during the duration of a log being enqueued

### DIFF
--- a/SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
+++ b/SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Azure.Mobile.Push
             }
             if (customData != null)
             {
-                PushNotificationReceived?.Invoke(null, new PushNotificationReceivedEventArgs()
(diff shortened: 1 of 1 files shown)


SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.UWP/Push.cs
```diff
@@ -81,14 +81,14 @@ private void ApplyEnabledState(bool enabled)
...
                                 MobileCenterLog.Debug(LogTag, $"Push token '{pushToken}'");
-                                var pushInstallationLog = new PushInstallationLog(null, null, pushToken, Guid.NewGuid());
-                                await Channel.EnqueueAsync(pushInstallationLog).ConfigureAwait(false);
                             }
                             else
                             {
                                 MobileCenterLog.Error(LogTag, "Push service registering with Mobile Center backend has failed.");
                             }
                         }
+                        var pushInstallationLog = new PushInstallationLog(null, null, pushToken, Guid.NewGuid());
```
guperrot: This could be executed after SDK disabled now, better fix would just be remove await as we don't need to wait here, await was just to disable the warning.

PR #371, opened 2017-06-30
ScreenSizeProvider unit tests added

bmourat:


SDK/MobileCenter/Microsoft.Azure.Mobile.UWP/Utils/DefaultScreenSizeProvider.cs
```diff
@@ -33,13 +33,13 @@ public DefaultScreenSizeProvider()
                 Task.Factory.StartNew(() =>
                 {
                     var displayInfo = DisplayInformation.GetForCurrentView();
-                    UpdateDisplayInformation(displayInfo, null);
+                    UpdateDisplayInformation((int)displayInfo.ScreenHeightInRawPixels, (int)displayInfo.ScreenWidthInRawPixels);
```
guperrot: why not `UpdateDisplayInformationHandler(displayInfo, null);` then

Tests/Microsoft.Azure.Mobile.Test.UWP/Utils/DefaultScreenSizeProviderTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        /// <summary>
+        /// Verify new screen size is set after screen size change 
+        /// </summary>
+        [TestMethod]
+        public void VerifyNewScreenSizeIsSet()
+        {
+            var newScreenHeight = 1;
+            var newScreenWidth = 1;
```
guperrot: change one of the dimension values so that they are not the same

SDK/MobileCenter/Microsoft.Azure.Mobile.UWP/Utils/DefaultScreenSizeProvider.cs
bmourat: It is not possible to create DisplayInformation object other than using
```DisplayInformation.GetForCurrentView()```
method, so to test size changed event i've made some refactoring.

guperrot: But why introducing `UpdateDisplayInformationHandler` ? Does not seem to be used anywhere now.

PR #381, opened 2017-07-10
Compatible with non XAML applications

MatkovIvan:
Used `Package.Current.Id.Name` instead of `Application.Current.GetType().Namespace` in `GetAppNamespace()` to not use `Windows.UI.Xaml.Application.Current` and be compatible with non XAML applications.
`app_namespace` description:
> The bundle identifier, package identifier, or namespace, depending on what the individual platforms use,  .e.g com.microsoft.example.

Also subscription to XAML unhandled exceptions wrapped into `try`/`catch` block to avoid fail on non XAML applications.

### DIFF
--- a/SDK/MobileCenter/Microsoft.Azure.Mobile.UWP/Utils/ApplicationLifecycleHelper.cs
+++ b/SDK/MobileCenter/Microsoft.Azure.Mobile.UWP/Utils/ApplicationLifecycleHelper.cs
@@ -51,14 +51,14 @@ namespace Microsoft.Azure.Mobile.Utils
             {
                 try
                 {
-                    // intentionally propagating exception to get the exception object that crashed the app.
+                    // Intentionally propagating exception to get the exception object that crashed the app.
                     eventArgs.UnhandledError.Propagate();
                 }
                 catch (Exception exception)
                 {
                     UnhandledExceptionOccurred?.Invoke(sender, new UnhandledExceptionOccurredEventArgs(exception));
 
-                    // If we don't throw exception - app will not be crashed. We need to throw to not change the app behavior.
+                    // Since UnhandledError.Propagate marks the error as Handled, rethrow in order to only Log and not Handle.
                     throw;
                 }
             };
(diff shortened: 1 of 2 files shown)


achocron:
@MatkovIvan I'm good with these changes, but is there a way to deal with unhandled exceptions in a D3D applications, or do they need to be ignored?

SDK/MobileCenter/Microsoft.Azure.Mobile.UWP/Utils/ApplicationLifecycleHelper.cs
```diff
@@ -46,17 +45,23 @@ public ApplicationLifecycleHelper()
...
-            }
+                    // intentionally propagating exception to get the exception object that crashed the app.
+                    eventArgs.UnhandledError.Propagate();
+                }
+                catch (Exception exception)
+                {
+                    UnhandledExceptionOccurred?.Invoke(sender, new UnhandledExceptionOccurredEventArgs(exception));
+
+                    // If we don't throw exception - app will not be crashed. We need to throw to not change the app behavior.
+                    throw;
```
achocron: According to the [docs](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.core.coreapplication):

> As part of an app's life cycle management code, developers can inspect error data (UnhandledError ) from an UnhandledErrorDetected event, and choose whether to mark the error as handled. If the error is marked handled in the event data, then execution can continue. If the error is not marked handled, the app and its process will be terminated.

So why re-throw? Also, do we want to raise the event if it has already been marked handled? Is that possible?

wind-rider:
@MatkovIvan will this PR bring support for Background Applications in Windows IoT Core (since they are XAML-less)?

MatkovIvan:
@wind-rider This PR target to support Direct3D XAML-less applications.

wind-rider:
@MatkovIvan ok, but will it support for Background UWP applications in Windows IoT Core as a side effect then?

MatkovIvan:
@wind-rider Feel free to try ;) It's quite possible. But keep in mind that this isn't target of this PR.

SDK/MobileCenter/Microsoft.Azure.Mobile.UWP/Utils/ApplicationLifecycleHelper.cs
MatkovIvan: > Since UnhandledError.Propagate marks the error as Handled, rethrow in order to only Log and not Handle
See (C++ but event the same): https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.core.icoreapplicationunhandlederror#events_

Also in example above `Handled` is checked, so if we don't check this we will log handled errors as well.
Updated comment for that to make it more clear.

PR #393, opened 2017-08-08
Feature/update bindings for new crash logic

achocron:


achocron:
The version being pulled in the script is not up to date, so don't try testing with that one

SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.iOS.Bindings/ApiDefinition.cs
```diff
@@ -214,38 +214,57 @@ interface MSStackFrame
 
     [Protocol, Model]
     [BaseType(typeof(NSObject))]
-    interface MSWrapperCrashesInitializationDelegate
-    {
-        //- (BOOL) setUpCrashHandlers;
-        [Export("setUpCrashHandlers")]
-        bool SetUpCrashHandlers();
+    interface MSCrashHandlerSetupDelegate
+   {
```
MatkovIvan: Formatting is broken a bit here

SDK/MobileCenterCrashes/Microsoft.Azure.Mobile.Crashes.iOS/ErrorReportCache.cs
```diff
@@ -1,23 +0,0 @@
-﻿using System.Collections.Generic;
```
guperrot: Why?

achocron: Because the disk instances of wrapper exception are cached in memory by native SDK, so there's no need to complicate logic by unnecessarily caching

PR #404, opened 2017-08-16
Explicit set C# language version

MatkovIvan:
Since CI does not support C# 7, this allows preventing misuse it locally - disable IDE suggestions and get compile error like:
`Feature 'local functions' is not available in C# 6.  Please use language version 7 or greater.`


MatkovIvan:
Oops, CI supports it now. Сlose it.

PR #417, opened 2017-08-22
Add push launched check overload with args as string

MatkovIvan:


MatkovIvan:
The build has been failed, because `The agent lost communication with the server.`

PR #422, opened 2017-08-29
Catch InvalidOperationException at startup as well

achocron:


SDK/MobileCenter/Microsoft.Azure.Mobile.UWP/Utils/ApplicationLifecycleHelper.cs
```diff
@@ -105,7 +105,12 @@ public ApplicationLifecycleHelper()
             }
             catch (COMException)
             {
-                // If MainView can't be accessed, a COMException is thrown. It means that the
+                // If MainView can't be accessed, a COMException or InvalidOperationException is thrown. It means that the
```
dhei: From a clarity perspective, it might be worth to log something at `Info` or `Debug` level to inform the developer about this exception instead of swallow it. What do you think?

achocron: We could. The reason we don't is that it isn't really an error. It simply means that the UI hasn't been set up yet. (That doesn't necessarily mean we shouldn't add a log though)

dhei: @achocron Ok, make sense. Logging a `Debug` level message should be enough I guess.

PR #432, opened 2017-09-05
Feature/test scripts

achocron:
Usage
Mac:
`./scripts/distribute-release.sh` (Inspect script for options)
`./scripts/send-push.sh` (Inspect script for options)
Windows:
`.\scripts\send-push.ps1` (Inspect script for options)

Note: You will need to set the following environment variables
- `MOBILE_CENTER_API_TOKEN`
- `MOBILE_CENTER_INT_API_TOKEN`

### DIFF
--- a/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.Droid/Contoso.Forms.Puppet.Droid.csproj
+++ b/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.Droid/Contoso.Forms.Puppet.Droid.csproj
@@ -31,7 +31,6 @@
     <WarningLevel>4</WarningLevel>
     <AndroidLinkMode>None</AndroidLinkMode>
     <EmbedAssembliesIntoApk>false</EmbedAssembliesIntoApk>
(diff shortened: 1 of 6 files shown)


Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.Droid/Contoso.Forms.Puppet.Droid.csproj
```diff
@@ -31,7 +31,6 @@
     <WarningLevel>4</WarningLevel>
     <AndroidLinkMode>None</AndroidLinkMode>
     <EmbedAssembliesIntoApk>false</EmbedAssembliesIntoApk>
-    <AndroidEnableMultiDex>true</AndroidEnableMultiDex>
```
guperrot: are you sure about that? debug has always been so huge with Firebase

achocron: It is actually still enabled, if you open the project in VS it will say multidex enabled. I think it's possible because it says multidex = true in the section above, which is for debug configurations of any platform

PR #434, opened 2017-09-07
Handled exceptions for Xamarin

guperrot:


Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/CrashesContentPage.xaml.cs
```diff
@@ -25,27 +25,41 @@ public CrashesContentPage()
...
         {
-            /* This is supposed to cause a crash, so we don't care that the variable 'x' is never used */
-#pragma warning disable CS0219
-            int x = (42 / int.Parse("0"));
-#pragma warning restore CS0219
+            try
+            {
+                action();
+            }
+            catch (Exception e)
```
achocron: I think this would be valid C# (and a bit neater):
```
try
{
    action();
}
catch (Exception e) when (HandleExceptionsSwitchCell.On)
{
    Crashes.TrackException(e);
}
```

dhei: @achocron I like your C# 6 exception filter approach! 👍

guperrot: Will get back to it once develop is fixed for cake.

guperrot: fixed

PR #437, opened 2017-09-14
Feature/new cake

achocron:
Update bootstrappers and add argument to make cake scripts compile

Update: Will troubleshoot CI on monday

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -4,17 +4,52 @@
 # Feel free to change this file to fit your needs.
 ##########################################################################
 
+<#
+
+.SYNOPSIS
+This is a Powershell script to bootstrap a Cake build.
+
+.DESCRIPTION
+This Powershell script will download NuGet if missing, restore NuGet tools (including Cake)
+and execute your Cake build script with the parameters you provide.
+
+.PARAMETER Script
+The build script to execute.
+.PARAMETER Target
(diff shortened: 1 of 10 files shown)


build.ps1
```diff
@@ -46,70 +81,65 @@ function MD5HashFile([string] $filePath)
...
 }
 
 # Make sure that packages.config exist.
 if (!(Test-Path $PACKAGES_CONFIG)) {
-    Write-Host -Message "Downloading packages.config..."
-    try { (New-Object System.Net.WebClient).DownloadFile("http://cakebuild.net/download/bootstrapper/packages", $PACKAGES_CONFIG) } catch {
+    Write-Verbose -Message "Downloading packages.config..."    
+    try {        
+        $wc = GetProxyEnabledWebClient
+        $wc.DownloadFile("https://cakebuild.net/download/bootstrapper/packages", $PACKAGES_CONFIG) } catch {
```
dhei: [styling] can you start `catch` in a newline?

achocron: I think we shouldn't modify this file for style as it was downloaded from cake's servers

achocron: only make necessary modifications. Otherwise we will be styling the file every time a new bootstrapped is available

PR #466, opened 2017-10-06
Fix calling Push.setEnabled(true) cancelling token

guperrot:


SDK/MobileCenterPush/Microsoft.Azure.Mobile.Push.Windows.Shared/Push.cs
```diff
@@ -89,10 +89,10 @@ public override bool InstanceEnabled
                 using (_mutex.GetLock())
                 {
                     var prevValue = InstanceEnabled;
-                    base.InstanceEnabled = value;
-                    _mutex.InvalidateState();
                     if (value != prevValue)
                     {
+                        base.InstanceEnabled = value;
```
achocron: Only thing I'd say is move `base.InstanceEnabled = value` back outside of the `if` statement so that the logs that are supposed to be printed are printed (e.g. "already enabled/disabled this service")

PR #506, opened 2017-11-16
Feature/rename in test scripts

achocron:
Don't merge until the "start ui tests" build finishes and passes (build # 1856)
1856 failed, but rebuilt and now it gets farther. Suspicious...

### DIFF
--- a/Tests/UITests/packages.config
+++ b/Tests/UITests/packages.config
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="NUnit" version="2.6.4" targetFramework="net45" />
-  <package id="Xamarin.UITest" version="2.0.3" targetFramework="net45" />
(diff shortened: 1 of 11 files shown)


Tests/UITests/packages.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="NUnit" version="2.6.4" targetFramework="net45" />
-  <package id="Xamarin.UITest" version="2.0.3" targetFramework="net45" />
+  <package id="Xamarin.UITest" version="2.2.1" targetFramework="net45" />
```
guperrot: isn't that the package that we had to pin version?

achocron: I don't remember that conversation. But now if we use 2.0.3 it fails and says to use the newer version.

guperrot:
@achocron are you running CI against this branch to test?

achocron:
@guperrot yeah i did and it failed the first time and passed the second

PR #508, opened 2017-11-16
Remove firebase dep

achocron:
Don't merge until firebase change is in official release for android sdk


Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/App.xaml.cs
```diff
@@ -46,22 +46,6 @@ public App()
             RealUserMeasurements.SetRumKey("b1919553367d44d8b0ae72594c74e0ff");
             AppCenter.Start($"uwp={UwpKey};android={AndroidKey};ios={IosKey}",
                                typeof(Analytics), typeof(Crashes), typeof(Distribute), typeof(Push), typeof(RealUserMeasurements));
-
```
guperrot: actually we will need to test all of that again not to break compat, the analytics enabled method will be restored

guperrot: we better either keep google-services.json and set analytics in forms puppet, or move that to the non forms puppet

Apps/Contoso.Android.Puppet/MainActivity.cs
```diff
@@ -49,12 +52,23 @@ protected override void OnCreate(Bundle savedInstanceState)
...
-            AppCenter.Start("bff0949b-7970-439d-9745-92cdc59b10fe", typeof(Analytics), typeof(Crashes), typeof(Distribute));
+
+            // Enable Firebase Analytics if set
+            var enableAnalytics = Preferences.SharedPreferences.GetBoolean(Constants.FirebaseAnalyticsEnabledKey, false);
+            if (enableAnalytics)
+            {
+                Push.EnableFirebaseAnalytics();
+            }
+
+            Push.PushNotificationReceived -= PrintNotification;
```
guperrot: Is this needed even when the method is static?
Should we investigate a bit more? Customers will hit this problem too and we don't document that.

achocron: I only did that because the explanation for doing it in a static constructor is to avoid duplicate handlers, and i didn't use a static constructor

guperrot: Ok, that's not what I meant.

If we really have an (ux) issue in our test app with that handler, there is a good chance customers will also have the issue, and it's either worth documenting or investigating to see if we can improve the consumption of this API.

Apps/Contoso.Android.Puppet/Resources/values/Strings.xml
```diff
@@ -17,6 +17,8 @@
 	<string name="AnalyticsTrackEvent">Track Event</string>
 	<string name="CrashesEnabled">Crashes Enabled</string>
 	<string name="TestCrash">Generate Test Crash</string>
+	<string name="PushEnabled">Push Enabled</string>
+    <string name="FirebaseAnalyticsEnabled">Firebase Analytics Enabled</string>
```
guperrot: please reformat the file

Apps/Contoso.Android.Puppet/packages.config
```diff
@@ -9,8 +9,17 @@
   <package id="Xamarin.Android.Support.Fragment" version="25.3.1" targetFramework="monoandroid71" />
   <package id="Xamarin.Android.Support.Media.Compat" version="25.3.1" targetFramework="monoandroid71" />
   <package id="Xamarin.Android.Support.Transition" version="25.3.1" targetFramework="monoandroid71" />
+  <package id="Xamarin.Android.Support.v4" version="25.3.1" targetFramework="monoandroid80" />
   <package id="Xamarin.Android.Support.v7.AppCompat" version="25.3.1" targetFramework="monoandroid71" />
   <package id="Xamarin.Android.Support.v7.RecyclerView" version="25.3.1" targetFramework="monoandroid71" />
   <package id="Xamarin.Android.Support.Vector.Drawable" version="25.3.1" targetFramework="monoandroid71" />
   <package id="Xamarin.Build.Download" version="0.4.4" targetFramework="monoandroid71" />
```
guperrot: please update targetFramework to monoandroid80 to avoid inconsistencies

Apps/Contoso.Android.Puppet/Properties/AndroidManifest.xml
```diff
@@ -1,5 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android" package="com.microsoft.appcenter.xamarin.puppet" android:versionCode="42" android:versionName="1.0.2-SNAPSHOT" android:installLocation="auto">
 	<uses-sdk android:minSdkVersion="16" />
-	<application android:label="SXPuppet" android:icon="@drawable/Icon" android:theme="@style/PuppetTheme" />
+	<application android:label="SXPuppet" android:icon="@drawable/Icon" android:theme="@style/PuppetTheme">
+		<receiver android:name="com.google.firebase.iid.FirebaseInstanceIdInternalReceiver" android:exported="false" />
```
guperrot: you also need to remove those from the forms.puppet one as we removed firebase in that one.
When the registration intent will be fired in the non firebase app it will try to launch FirebaseInstanceIdReceiver and class not found.

guperrot: Fixed in https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/508/commits/8c218f6c059bb1f7150729fa143199c9383452d3

PR #519, opened 2017-11-21
Remove "crashes are not supported" log on windows

MatkovIvan:


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -28,7 +28,6 @@ public static Crashes Instance
 
         public void OnChannelGroupReady(IChannelGroup channelGroup, string appSecret)
         {
-            AppCenterLog.Warn(AppCenterLog.LogTag, "Crashes service is not yet supported on this platform.");
```
guperrot: We also need to surface registration errors, there is a if debug to remove to log registration exception.

PR #526, opened 2017-11-29
Feature/organize build scripts

achocron:
CI tested and working


scripts/configuration/AppCenterModule.cake
```diff
@@ -0,0 +1,48 @@
+// AppCenter module class definition.
+class AppCenterModule
+{
+    public string AndroidModule { get; }
```
guperrot: Isn't that dead code the Android one?

scripts/configuration/ac-build-config.xml
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0"?>
+<config>
+    <sdkInfo>
+        <sdkVersion>1.0.2-SNAPSHOT</sdkVersion>
+        <iosVersion>1.0.0</iosVersion>
+        <androidVersion>1.0.0</androidVersion>
```
guperrot: what is that file and how versions are maintained here?

scripts/configuration/AppCenterModule.cake
achocron: wow you're right

scripts/configuration/ac-build-config.xml
achocron: iosVersion and androidVersion are the versions that we bind; sdkVersion is the version that we ship. The scripts are updated to change that value and read it instead of using a csproj file

scripts/configuration/AppCenterModule.cake
```diff
@@ -0,0 +1,45 @@
...
+
+    private AppCenterModule(XmlNode moduleNode, string nuspecFolder, string nugetVersion)
+    {
+        DotNetModule = moduleNode.Attributes.GetNamedItem("dotnetModule").Value;
+        MainNuspecFilename = moduleNode.Attributes.GetNamedItem("nuspec").Value;
+        NuGetVersion = nugetVersion;
+        var nuspecText = Statics.Context.FileReadText(System.IO.Path.Combine(nuspecFolder, MainNuspecFilename));
+        var startTag = "<id>";
+        var endTag = "</id>";
+        int startIndex = nuspecText.IndexOf(startTag) + startTag.Length;
```
guperrot: Why not using xml apis then?

scripts/configuration/AssemblyGroup.cake
```diff
@@ -0,0 +1,62 @@
...
+        }
+        return groups;
+    }
+
+    public string NuspecKey => $"${Id}_dir$";
+    public string Id { get; set; }
+    public string Folder { get; set; }
+    public IList<string> AssemblyPaths { get; set; }
+    public IList<AssemblyGroup> Subgroups { get; set; }
+    public bool Download { get; set; }
```
guperrot: blank lines missing here

scripts/configuration/AppCenterModule.cake
achocron: This is older code that i just moved, but for now I think it's probably simple enough to keep. (We do similar things in version.cake)

jwallra: But they're used elsewhere in this file.... I would favor consistency within a file over consistency across files

PR #530, opened 2017-12-05
Fix exception when unhandled exception occurs after shutdown

MatkovIvan:


achocron:
Sorry for the constant closing/reopening, CI is not starting builds correctly.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -292,6 +294,18 @@ private void SetInstanceCustomProperties(CustomProperties customProperties)
             _channel.EnqueueAsync(customPropertiesLog);
         }
 
+        private void OnUnhandledExceptionOccurred(object sender, UnhandledExceptionOccurredEventArgs args)
+        {
+            lock (AppCenterLock)
+            {
+                if (_channelGroup != null)
+                {
+                    _channelGroup.ShutdownAsync();
```
achocron: Can you add a debug log that says "Shutting down App Center due to unhandled exception"?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Channel/ChannelGroup.cs
```diff
@@ -99,6 +100,15 @@ public void SetEnabled(bool enabled)
 
         public async Task ShutdownAsync()
         {
+            lock (_channelGroupLock)
```
achocron: Just out of curiosity, why is this before the call to `ThrowIfDisposed()`?

MatkovIvan: Doesn't matter I guess. Will change, not big deal

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
MatkovIvan: "Shutting down **channel group** due to unhandled exception."

PR #533, opened 2017-12-06
Bug: Use device locale instead of App Locale

jwallra:
Not very happy with this solution but it seems pretty accepted as the way to always get the device/system locale.  See link in the helper class.

I think, though, that if the implementation of the DLL changes this may break, which isn't great.

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
@@ -1,0 +1,50 @@
+﻿using System.Globalization;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Microsoft.AppCenter.Windows.Shared.Utils
+{
+    //https://www.pedrolamas.com/2017/05/03/cultureinfo-changes-in-uwp-part-2/
+    class CultureInfoHelper
+    {
+        private const uint LOCALE_SNAME = 0x0000005c;
+        private const string LOCALE_NAME_USER_DEFAULT = null;
+        private const string LOCALE_NAME_SYSTEM_DEFAULT = "!x-sys-default-locale";
+        private const int BUFFER_SIZE = 530;
+        
+        [DllImport("api-ms-win-core-localization-l1-2-0.dll", CharSet = CharSet.Unicode)]
+        private static extern int GetLocaleInfoEx(string lpLocaleName, uint LCType, StringBuilder lpLCData, int cchData);
+
+        public static CultureInfo GetCurrentCulture()
+        {
+            var name = InvokeGetLocaleInfoEx(LOCALE_NAME_USER_DEFAULT, LOCALE_SNAME);
+
+            if (name == null)
+            {
+                name = InvokeGetLocaleInfoEx(LOCALE_NAME_SYSTEM_DEFAULT, LOCALE_SNAME);
+
+                if (name == null)
+                {
+                    // If system default doesn't work, use invariant
+                    return CultureInfo.InvariantCulture;
+                }
+            }
+
+            return new CultureInfo(name);
+        }
+
+        private static string InvokeGetLocaleInfoEx(string lpLocaleName, uint LCType)
+        {
+            var buffer = new StringBuilder(BUFFER_SIZE);
+
+            var resultCode = GetLocaleInfoEx(lpLocaleName, LCType, buffer, BUFFER_SIZE);
+
+            if (resultCode > 0)
+            {
+                return buffer.ToString();
+            }
+
+            return null;
+        }
+    }
+}
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
@@ -4,13 +4,13 @@ using System.Text;
 
 namespace Microsoft.AppCenter.Windows.Shared.Utils
 {
-    //https://www.pedrolamas.com/2017/05/03/cultureinfo-changes-in-uwp-part-2/
+    // https://www.pedrolamas.com/2017/05/03/cultureinfo-changes-in-uwp-part-2/
     class CultureInfoHelper
     {
-        // relative location of system locale name
+        // Relative location of system locale name.
         private const uint LOCALE_SNAME = 0x0000005c;
 
-        // Constants used by the National Language Support reference
+        // Constants used by the National Language Support reference.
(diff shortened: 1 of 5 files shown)


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+        // relative location of system locale name
+        private const uint LOCALE_SNAME = 0x0000005c;
+
+        // Constants used by the National Language Support reference
+        private const string LOCALE_NAME_USER_DEFAULT = null;
+        private const string LOCALE_NAME_SYSTEM_DEFAULT = "!x-sys-default-locale";
+
+        private const int BUFFER_SIZE = 530;
+        
+        [DllImport("api-ms-win-core-localization-l1-2-0.dll", CharSet = CharSet.Unicode)]
```
achocron: Will using this dllimport cause any problems with windows store certification? I think that we should test that for UWP

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Globalization;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Microsoft.AppCenter.Windows.Shared.Utils
+{
+    //https://www.pedrolamas.com/2017/05/03/cultureinfo-changes-in-uwp-part-2/
+    class CultureInfoHelper
+    {
+        // relative location of system locale name
```
achocron: This is a nitpick, but r->R and we usually end comments with a period

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        private const int BUFFER_SIZE = 530;
+        
+        [DllImport("api-ms-win-core-localization-l1-2-0.dll", CharSet = CharSet.Unicode)]
+        private static extern int GetLocaleInfoEx(string lpLocaleName, uint LCType, StringBuilder lpLCData, int cchData);
+
+        public static CultureInfo GetCurrentCulture()
+        {
+            var name = InvokeGetLocaleInfoEx(LOCALE_NAME_USER_DEFAULT, LOCALE_SNAME);
+
```
achocron: extra line

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (name == null)
+            {
+                name = InvokeGetLocaleInfoEx(LOCALE_NAME_SYSTEM_DEFAULT, LOCALE_SNAME);
+
```
achocron: extra line

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+                if (name == null)
+                {
+                    // If system default doesn't work, use invariant
+                    return CultureInfo.InvariantCulture;
+                }
+            }
+
```
achocron: extra line

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+            return new CultureInfo(name);
+        }
+
+        private static string InvokeGetLocaleInfoEx(string lpLocaleName, uint LCType)
+        {
+            var buffer = new StringBuilder(BUFFER_SIZE);
+            var resultCode = GetLocaleInfoEx(lpLocaleName, LCType, buffer, BUFFER_SIZE);
+
```
achocron: extra line

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (resultCode > 0)
+            {
+                return buffer.ToString();
+            }
+
```
achocron: extra line

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/packages.config
```diff
@@ -15,7 +15,7 @@
   <package id="System.Diagnostics.Debug" version="4.0.11" targetFramework="net452" />
   <package id="System.Diagnostics.Tools" version="4.0.1" targetFramework="net452" />
   <package id="System.Diagnostics.Tracing" version="4.1.0" targetFramework="net452" />
-  <package id="System.Globalization" version="4.0.11" targetFramework="net452" />
```
guperrot: @jwallra @achocron how will that impact customers? Do we need to update nuspec dependencies?

guperrot:
Tested, PC UWP, phone UWP, PC WPF, PC winforms. It works.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
jwallra: Tested store certification; it failed against both develop and my branch, but both gave the same results, so I don't think there's any _change_ to certifiability at the very least.

achocron: Can you test certification in a vanilla app that uses the updated sdk? I know puppet doesn't pass, but a vanilla app might. I know there was no diff, but I think that the verification tool stops early when it detects failures, so the diff might not give the whole story here

jwallra: Interestingly, a blank app using the same modules as the puppet app has a few more errors in unmanaged APIs... but the other warnings are the same, and the errors from the blank app are a strict superset of the errors given by the puppet app.  I'd say we're ok though.

jwallra: Passed after changing to release - looks good!

PR #534, opened 2017-12-07
Changing HTTP exception message to match log message

jwallra:


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Ingestion/Http/HttpNetworkAdapter.cs
```diff
@@ -67,10 +67,11 @@ private HttpClient HttpClient
                 {
                     logPayload = "<binary>";
                 }
-                AppCenterLog.Verbose(AppCenterLog.LogTag, $"HTTP response status={(int)response.StatusCode} ({response.StatusCode}) payload={logPayload}");
+                String logMessage = $"HTTP response status={(int)response.StatusCode} ({response.StatusCode}) payload={logPayload}";
```
guperrot: we use var for local variables where possible

PR #553, opened 2017-12-12
Updating TPN with code from device locale fix

jwallra:


guperrot:
@jwallra was that reviewed by our CELA contacts?

jwallra:
Yes, I confirmed this is correct with Ariel Patano

PR #554, opened 2017-12-12
Revert "Apply externals command in bindings for all configurations"

achocron:
Reverts Microsoft/AppCenter-SDK-DotNet#552


guperrot:
@MatkovIvan https://asgard-int.trafficmanager.net/orgs/appcenter-sdk/apps/xamarin-puppet-android-03/build/branches/develop/builds/46 limitation on App Center build. We run externals as appcenter-post-clone instead to make it work.

If you have a solution that works in App Center build, you can resubmit a new PR.

PR #558, opened 2017-12-18
Fix OS version information on old windows

MatkovIvan:


guperrot:
@MatkovIvan we support only Windows 10 in App Center, do we need this?

PR #559, opened 2017-12-20
Fix/ui tests

guperrot:


guperrot:
This goes with a build definition change.
Android package unsigned built by the shell script was working fine until December 14th, but now it produces a corrupted android build (triggering a null reference in xamarin forms init) when built on VSTS.

This patch allows to run UI tests without cleaning and without rebuilding apps on VSTS and use VSTS tasks for everything else.

PR #561, opened 2017-12-21
Fix disabled state issues

MatkovIvan:
There are a couple of issues related to the disabled state.
- Core channel doesn't disabled when `AppCenter` was disabled in the previous session;
- Start services log send even if `AppCenter` was disabled;
- Channel **never** discard logs, even if `_discardLogs` flag is set;

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
@@ -274,9 +274,13 @@ namespace Microsoft.AppCenter
                     return;
                 }
 
+                // Update channels state.
                 _channelGroup?.SetEnabled(value);
+
+                // Store state in the application settings.
                 _applicationSettings.SetValue(EnabledKey, value);
 
+                // Apply change to services.
                 foreach (var service in _services)
                 {
                     service.InstanceEnabled = value;
@@ -337,6 +341,7 @@ namespace Microsoft.AppCenter
             _channelGroup = _channelGroupFactory?.CreateChannelGroup(_appSecret, _networkStateAdapter) ?? new ChannelGroup(_appSecret, null, _networkStateAdapter);
             _channel = _channelGroup.AddChannel(ChannelName, Constants.DefaultTriggerCount, Constants.DefaultTriggerInterval,
                                                 Constants.DefaultTriggerMaxParallelRequests);
+            _channel.SetEnabled(InstanceEnabled);
             if (_logUrl != null)
             {
                 _channelGroup.SetLogUrl(_logUrl);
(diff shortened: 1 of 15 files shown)


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -114,22 +116,56 @@ static Task<bool> PlatformIsEnabledAsync()
...
         {
             lock (AppCenterLock)
             {
-                Instance.InstanceEnabled = enabled;
-                return Task.FromResult(default(object));
+                var taskSource = new TaskCompletionSource<object>();
+                if (!Configured)
+                {
+                    AppCenterLog.Error(AppCenterLog.LogTag, NotConfiguredMessage);
+                    taskSource.SetException(new AppCenterException(NotConfiguredMessage));
```
guperrot: our sdk never throws exception and should not start doing it as we don't want to crash prod apps unexpectedly, please use only error logs.

Also if it gives the right result before Configure we can leave the behavior, I think iOS works too. Android is different and might be improved.

MatkovIvan: @guperrot Yeah, I'll revert this. Also, the most of the current problems with tests caused by this.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -335,7 +299,8 @@ private void SetInstanceCustomProperties(CustomProperties customProperties)
         {
             if (!Configured)
             {
-                AppCenterLog.Error(AppCenterLog.LogTag, NotConfiguredMessage);
+                AppCenterLog.Error(AppCenterLog.LogTag, "App Center hasn't been configured. " +
```
guperrot: If resharper told you to inline this as only 1 occurrence, we should not follow the advice as we may add new calls in the future.

Tests/Microsoft.AppCenter.Test.Windows/AppCenterTest.cs
```diff
@@ -300,13 +308,26 @@ public void StartWrongInstanceTypeServiceAndCorrectService()
...
+
+            AppCenter.Configure("some string");
+
+            Assert.IsTrue(AppCenter.Configured);
+            _channelMock.Verify(channel => channel.SetEnabled(It.IsAny<bool>()));
+        }
 
         /// <summary>
-        /// Verify that starting a App Center instance with a null app secret does not cause App Centerto be configured
+        /// Verify that starting a AppCenter instance with a null app secret does not cause AppCenter to be configured
```
achocron: AppCenter -> App Center

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
MatkovIvan: @guperrot This is just revert. My change was vice versa :grin:

guperrot: Can you restore the constant in the message? It was nice change.

PR #563, opened 2017-12-25
Sending start services log after enable

MatkovIvan:


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -425,6 +433,26 @@ private void StartService(IAppCenterService service)
             AppCenterLog.Info(AppCenterLog.LogTag, $"'{service.GetType().Name}' service started.");
         }
 
+        private async Task SendStartServiceLog(IEnumerable<string> serviceNames)
+        {
+            if (InstanceEnabled)
```
guperrot: This check is not atomic. On Android we run commands in a loop so there is only 1 thread involved, but in .NET you have multiple threads involved when using tasks.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -405,7 +401,18 @@ internal void StartInstance(params Type[] services)
             // Enqueue a log indicating which services have been initialized
             if (serviceNames.Count > 0)
             {
-                Task.Run(async () => await SendStartServiceLog(serviceNames));
+                if (InstanceEnabled)
```
guperrot: not sure to follow the fix for potential race conditions by reading github diffs.

Is every code inside the lock statements?

guperrot: reading entire file it looks like it's okay now, will test again on puppet

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -395,9 +399,20 @@ internal void StartInstance(params Type[] services)
             }
 
             // Enqueue a log indicating which services have been initialized
-            if (startServiceLog.Services.Count > 0)
+            if (serviceNames.Count > 0)
             {
-                _channel.EnqueueAsync(startServiceLog);
+                if (InstanceEnabled)
+                {
+                    Task.Run(async () => await _channel.EnqueueAsync(new StartServiceLog { Services = serviceNames }).ConfigureAwait(false));
```
guperrot: The state can change by the time you enter the run code and execute enqueueAsync.
In Rum we handle that case with a cancellation token.
Or you can simply use the lock in the task and check again for InstanceEnabled.
I am not sure why you use a task.run in the first place in this method as we don't do that for example for custom properties.

MatkovIvan: `Task.Run` used to force run in the background and `_channel` will be disabled on state change as well, so that will be checked inside. But you are right, here it isn't necessary.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -263,7 +263,7 @@ private AppCenter()
 
         private bool InstanceEnabled => _applicationSettings.GetValue(EnabledKey, true);
 
-        private Task SetInstanceEnabledAsync(bool value)
+        private Task SetInstanceEnabled(bool value)
```
guperrot: It could be in future and also it returns a task and thus can be called with await.
If await works it should have async in the name? Also the PCL public is async even if ios uses Task.fromResult for now, but ideally all disk accesses should be async in the future.

@achocron thougts?

MatkovIvan: @guperrot This change caused by https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/566#discussion_r160217877 and addressed to avoid misreading with locks

achocron: It's in a confusing state right now since it returns a task but doesn't do async work i guess? I don't think that just returning a `Task` should make it `async` but if it is going to be `async` in the future maybe we should make it that way now to avoid conflicts in the future

guperrot: then I would consider building the task from result outside the method

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -261,34 +261,38 @@ private AppCenter()
...
+                service.InstanceEnabled = value;
             }
+            AppCenterLog.Info(AppCenterLog.LogTag, $"App Center has been {enabledTerm}.");
+
+            // Send started services.
+            if (_startedServiceNames != null && value)
+            {
+                var startServiceLog = new StartServiceLog { Services = _startedServiceNames };
+                _startedServiceNames = null;
+                return _channel.EnqueueAsync(startServiceLog);
```
MatkovIvan: @guperrot not from result here

guperrot: Hum, better add a comment then, it's weird to have a non async method return a task

PR #566, opened 2018-01-02
PR 563 mirror

guperrot:


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/AppCenter.cs
```diff
@@ -124,8 +125,7 @@ static Task PlatformSetEnabledAsync(bool enabled)
         {
             lock (AppCenterLock)
             {
-                Instance.InstanceEnabled = enabled;
-                return Task.FromResult(default(object));
+                return Instance.SetInstanceEnabledAsync(enabled);
```
achocron: Does the lock really have an effect since the operation inside the block is asynchronous? The `Task` will continue running potentially after the lock is released

guperrot: @achocron https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/563#discussion_r160085901

guperrot:
PR build from forks are now re-enabled, no need for mirrors anymore 💃

Issue #362, opened 2017-06-26
Support automatically report of country code in the UWP SDK


OpportunityLiu:
Can we support getting this info automatically?
What about `new Windows.Globalization.GeographicRegion().CodeTwoLetter`?
It reports the user's home region.
https://docs.microsoft.com/en-us/uwp/api/Windows.Globalization.GeographicRegion

guperrot:
Hi, the country code is meant for geographical position, not culture/locale.

On Analytics we already report the culture/locale automatically and you can see it on the portal in Audience > Analytics > Languages.

Countries chart on the left is meant for actual geographical position.
On Android and iOS, we report country only if you have a mobile carrier at the moment.

The API you mentioned cannot be used to determine geographical position.

guperrot:
> The API you mentioned cannot be used to determine geographical position, it would return the same result as device language settings.

It's even worse, it returns the language that the app supports (in manifest or resources if using x-generate) closest to the user system locale. Anyway this has nothing to do with actual geographical position.

michael-hawker:
Any updates on this? I had a crash in my app as the docs didn't have a sample for this and the example I got elsewhere was invalid.

Is it possible to get the user's location for UWP without adding a capability to the app?  It does seem like the Globalization API is going to be a rough approximation as I'd imagine most users who setup their PC in their own country would have this value set to their country?  Are there stats on the changes people make to this setting and how often it differs from the real location?

michael-hawker:
Also, according to the [location](https://docs.microsoft.com/en-us/windows/uwp/maps-and-location/get-location) docs that value can only be retrieved when the app is in the foreground, so the timing with when to set and start the App Center API would need to be carefully considered/documented.

AnastasiaKubova:
Hi there! We already have an `AppCenter.SetCountryCode` API for setting country code manually.  Please [see a code sample](https://docs.microsoft.com/en-us/appcenter/sdk/analytics/windows#country-code) for how to do it.
I'm closing this issue but feel free to reopen it if you will have additional questions.

PR #577, opened 2018-01-22
Use geographic region as country code source

MatkovIvan:
`RegionInfo.CurrentRegion` uses `CultureInfo.CurrentCulture` as the basis for the region it returns.

The home region is usually set when Windows is installed and is available in
**Settings → Time & language → Region & language → Country or region**.
`GeographicRegion` allows to get this value.


guperrot:
As discussed with @blparr and https://github.com/Microsoft/AppCenter-SDK-DotNet/issues/362#issuecomment-311197646 this does not reflect actual position and should not be automatically reported. Users can use that if they want unreliable data but should not be the default. Please rollback the latest commit, we can accept the first one which touch only test apps.

PR #579, opened 2018-01-23
Don't create extra channel in ThrowStorageExceptionInDeleteLogsTime test

bmourat:


Tests/Microsoft.AppCenter.Test.Windows/Channel/ChannelTest.cs
```diff
@@ -331,11 +333,14 @@ public void IngestionClosedOnNonRecoverableHttpError()
 
         private void SetChannelWithTimeSpan(TimeSpan timeSpan)
```
achocron: Can you instead add an optional argument for whether to create the channel? And add a comment for why it is needed?

guperrot: Why not just call `MakeIngestionCallsSucceed();` directly in that test instead of modifying the function and adding a context.

PR #581, opened 2018-01-23
Rename trackException to trackError in public surface

guperrot:


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared/Crashes.cs
```diff
@@ -159,13 +159,13 @@ public static void GenerateTestCrash()
         }
 
         ///// <summary>
-        ///// Track an exception. TODO make it public once backend supports it.
+        ///// Track an handled error. TODO make it public once backend supports it.
```
TroubleMakerBen: It should say "a handled error".

Issue #517, opened 2017-11-19
Sometimes App hangs without internet connection UWP SDK


snrnats:
I am using MobileCenter 0.17.1 in the App. I've got a lot of hangs in Developer Dashboard. I suspect MobileCenter.Start method in App.ctor().
[devCenter.zip](https://github.com/Microsoft/AppCenter-SDK-DotNet/files/1485842/devCenter.zip)

I was able to reproduce the hang in local environment both MobileCenter 0.17.1 and AppCenter 1.0.1. It happens rarely. Steps:
- Disable internet on the target device.
- Run a in-process background task without launching the app.
- It hangs on PauseExecutionUntilNetworkBecomesAvailable()
-- I also tried disable/enable internet connection while running background task. It seems to cause issue much often.
[appcenterlog.txt](https://github.com/Microsoft/AppCenter-SDK-DotNet/files/1485852/appcenter.txt)
[LocalDump](https://1drv.ms/u/s!AmwrM5g-csqJk9dcQHdAtLxcywOiMA)
![untitled](https://user-images.githubusercontent.com/1597284/32994081-67ff882c-cd73-11e7-8d10-c2422bb0c905.png)

achocron:
Thanks! Added to our backlog.

pfaucon:
Looking to start using App Center once managed exceptions are supported, is there any update on this ?

elamalani:
@pfaucon Managed exceptions will be supported in App Center for Xamarin platform very shortly (most likely in a week or two). If you want to get early preview access to that feature, please share your email address that you use for App Center.

elamalani:
@pfaucon We have prioritized this work in our backlog and will fix it soon.

EmilAlipiev:
is there any progress on this issue?

elamalani:
@EmilAlipiev There is no update on this bug yet. We will let you know once this is fixed and ready for a release.

MatkovIvan:
Hi @snrnats 

The fix for this is merged to `develop`, but not released yet.
It would be great if you could verify that this issue is solved now.

Best,
Ivan

snrnats:
Hi @MatkovIvan 

I wasn't able to reproduce the bug even using the reported version 1.0.1. Seems like something has changed on my side since then.

guperrot:
@snrnats [1.6.0](https://github.com/Microsoft/AppCenter-SDK-DotNet/releases/tag/1.6.0) has been released with both your PRs.

PR #602, opened 2018-02-19
[Windows] Unwrap AggregateException when catching exceptions from Task.Wait() method

snrnats:
We need to unwrap AggregateException when using Task.Wait() method. So that the caller will get original exception.


guperrot:
Hi, we are not sure what you are trying to improve here.
Can you tell us more about the intent of this change?

snrnats:
In all cases `Wait` method is called on tasks that can throw `StorageException`. But `Wait` call will wrap up this exception with `AggregateException` [according to docs](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.wait?view=netframework-4.7.1#System_Threading_Tasks_Task_Wait) .

Here is the issue: the caller of `PutLog` method expects `StorageException` type of exception and tries to catch it (see `Changel.PersistLogAsync()`), but it won’t happen because AggregateException will be thrown and no code will handle it.

I changed all other `Wait` calls for consistency and to prevent further issues.

guperrot:
Hi, I mean, what does this change achieve for real? How is user experience impacted and what are the motivations for this?

snrnats:
My app is affected by the issues I described. I've got some crash reports when using AppCenter.Push 1.3.
```
Application Specific Information:
TaskExceptionHolder_UnhandledException (AggregateException_ctor_DefaultMessage (The storage operation failed))

Exception Stack:
unknown location
SharedLibrary!<BaseAddress>+0x4eb21d
System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, Threading.CancellationToken cancellationToken)
System.Threading.Tasks.Task.Wait()
Microsoft.AppCenter.Storage.Storage.<>c__DisplayClass12_0.<PutLog>b__0()
System.Action.Invoke()
SharedLibrary!<BaseAddress>+0x4f52e5
SharedLibrary!<BaseAddress>+0x4f520c
System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
```

```
SQLite.SQLiteException: Busy
Microsoft.AppCenter.Storage.StorageException: The storage operation failed
System.AggregateException: AggregateException_ctor_DefaultMessage (The storage operation failed)
System.AggregateException: TaskExceptionHolder_UnhandledException (AggregateException_ctor_DefaultMessage (The storage operation failed))
```

guperrot:
Do you have a reproduction scenario for this crash?

snrnats:
No, i don't have steps. It reproduces not very often according to crash count. For now I can tell that the crash occurs only while executing code from in process background task.

guperrot:
But how would that patch solve a crash that we don't even reproduce?

snrnats:
My crash logs say that app crashes because `AggregateException `isn't handled in the code. In our case the `AggregateException `contains `StorageException`. Try-catch for `StorageException `is already presented in 
 `Changel.PersistLogAsync()`. We just need to unwrap the exception.

It will prevent the crash. Although I am not sure about the reason of underlying `SQLite.SQLiteException: Busy`

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -52,7 +53,7 @@ public Storage() : this(DefaultAdapter())
         internal Storage(IStorageAdapter adapter)
         {
             _storageAdapter = adapter;
-            _queue.Add(new Task(() => InitializeDatabaseAsync().Wait()));
```
guperrot: Like @MatkovIvan mentioned in a conversation we had, just using `.GetAwaiter().GetResult()`  would remove the need for an extension utils, please switch to that and remove the additional extension.

snrnats:
Hi @guperrot 
I will try to add unit test a little bit later.

Tests/Microsoft.AppCenter.Test.Windows/Storage/FakeStorageTest.cs
```diff
@@ -85,7 +85,7 @@ public void GetLogsQueryError()
             var mockAdapter = new Mock<IStorageAdapter>();
             mockAdapter.Setup(
                     a => a.GetAsync(It.IsAny<PredType>(), It.IsAny<int>()))
-                .Throws(new StorageException());
+                .ThrowsAsync(new StorageException());
```
snrnats: `ThrowsAsync()` allows to setup mock that returns faulted task. `Throws()` throws right away without returning a task.

Tests/Microsoft.AppCenter.Test.Windows/TaskExtension.cs
```diff
@@ -8,26 +8,11 @@ public static class TaskExtension
     {
```
guperrot: Hi, thanks for the changes. However since we use GetAwaiter().GetResult() directly in code, we should remove this utility.

snrnats:
Hi. I noticed strange behavior of unit tests. Running all tests complete successfully, but if I filter tests with "Storage" and run them then `ThrowStorageExceptionInDeleteLogsTime` will fail.

```
Debug Trace:
[AppCenter] WARN: Channel is temporarily disabled; log was saved to disk
[AppCenter] DEBUG: CheckPendingLogs(channelName) pending log count: 1
[AppCenter] DEBUG: triggerIngestion(channelName) pendingLogCount=1
[AppCenter] DEBUG: CheckPendingLogs(channelName) pending log count: 0
[AppCenter] WARN: Could not delete logs for batch test-batch-id
Microsoft.AppCenter.Storage.StorageException: The storage operation failed
   at Moq.MethodCall.Execute(ICallContext call)
   at Moq.MethodCallReturn`2.Execute(ICallContext call)
   at Moq.ExecuteCall.HandleIntercept(ICallContext invocation, InterceptorContext ctx, CurrentInterceptContext localctx)
   at Moq.Interceptor.Intercept(ICallContext invocation)
   at Moq.Proxy.CastleProxyFactory.Interceptor.Intercept(IInvocation invocation)
   at Castle.DynamicProxy.AbstractInvocation.Proceed()
   at Castle.Proxies.ObjectProxy_1.DeleteLogs(String channelName, String batchId)
   at Microsoft.AppCenter.Channel.Channel.HandleSendingSuccess(State state, String batchId) in D:\Work\AppCenter-SDK-DotNet\SDK\AppCenter\Microsoft.AppCenter.Windows.Shared\Channel\Channel.cs:line 432
```

Seems like 'HandleSendingSuccess' needs to await for `_storage.DeleteLogs(Name, batchId)`, but even then the test will fail. Do you observe the same behavior?

devenv_2018-02-23_10-43-04

snrnats:
Waiting for a next release of moq4 where [an issue](https://github.com/moq/moq4/issues/592) affecting this PR will be resolved

guperrot:
Hi @snrnats, are you still blocked on moq release?

snrnats:
Hi @guperrot . Yes, still waiting for the next moq4 release.

I'm trying to reach out to authors of the moq4, hope they will give us more information about the next release

guperrot:
You will also need to fix the conflicts, a big refactoring was introduced to address #517.

Tests/Microsoft.AppCenter.Test.Windows/Storage/FakeStorageTest.cs
```diff
@@ -85,11 +85,11 @@ public void GetLogsQueryError()
...
```
MatkovIvan: You can use
```C#
.Throws(TaskExtension.GetFaultedTask(new StorageException()))
```
to avoid moq bug.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageTest.cs
```diff
@@ -24,7 +24,7 @@ public void InitializeStorageTest()
         [TestMethod]
         public void CountEmptyStorage()
         {
-            var count = _storage.CountLogsAsync(StorageTestChannelName).RunNotAsync();
+            var count = _storage.CountLogsAsync(StorageTestChannelName).GetAwaiter().GetResult();
```
MatkovIvan: Tests can be `async` and I've already replaced it partially.
`async`/`await` looks like right replacement instead of `GetAwaiter().GetResult()`
You don't necessarily change all the tests in the project, only those that relate to your fix.

Tests/Microsoft.AppCenter.Test.Windows/Storage/FakeStorageTest.cs
snrnats: This line of code produce an error `Error CS1503	Argument 1: cannot convert from 'System.Threading.Tasks.Task<Microsoft.AppCenter.Storage.StorageException>' to 'System.Exception'`.

Anyway we can't use `Throws()` to mock an exception in async method.
I think that there is an another way to rewrite it. We could use `Returns()` with a faulted task. It will look bulky but will do what was intended.
`.Returns(TaskExtension.GetFaultedTask<List<Microsoft.AppCenter.Storage.Storage.LogEntry>>(new StorageException()))`

I can try to change it in a such way if we don't want to wait for a moq4 release.

snrnats:
I reverted changes that are not related to fix

Tests/Microsoft.AppCenter.Test.Windows/Storage/FakeStorageTest.cs
MatkovIvan: Oops, I mean `Returns` the faulted task of course. You understood my idea correctly :+1:

PR #605, opened 2018-02-27
Add readme.txt file to be included in the NuGet

elamalani:
Add readme.txt file to be included in the NuGet package which opens up after any public App Center Nuget is installed in an application.

### DIFF
--- a/nuget/readme.txt
+++ b/nuget/readme.txt
@@ -1,6 +1,6 @@
--------------------------------------------
+------------------------------
 App Center SDK for .NET Readme
--------------------------------------------
+------------------------------
 
 Follow the “Getting Started” link to set up your application with App Center SDK. Then click on the service link for each App Center service that you want to enable in your app.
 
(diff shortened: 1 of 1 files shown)


nuget/readme.txt
```diff
@@ -0,0 +1,24 @@
+-------------------------------------------
+App Center SDK for .NET Readme
+-------------------------------------------
+
+Follow the “Getting Started” link to set up your application with App Center SDK. Then click on the service link for each App Center service that you want to enable in your app.
+
+-----Getting Started-----
+Xamarin: https://docs.microsoft.com/en-us/appcenter/sdk/getting-started/xamarin
```
guperrot: blank lines after the header should be better

nuget/readme.txt
```diff
@@ -0,0 +1,24 @@
+-------------------------------------------
+App Center SDK for .NET Readme
+-------------------------------------------
+
+Follow the “Getting Started” link to set up your application with App Center SDK. Then click on the service link for each App Center service that you want to enable in your app.
+
+-----Getting Started-----
```
guperrot: should have same length of ---- for all headers

nuget/readme.txt
elamalani: Not sure what should be changed here?

guperrot: blank lines after each header section

guperrot:
@elamalani edited myself

PR #607, opened 2018-02-27
Merging develop into master

nikitag-ms:


guperrot:
Conflicts plus not ready to do it, as I explained Michael, there is some work to do before releasing to bring native changes to xamarin.

PR #614, opened 2018-03-03
Fix forms preview in our own apps

guperrot:


guperrot:
@rizamarhaban there are known issues in Firebase intents when swiping the application (something we have no control on) : https://github.com/firebase/quickstart-android/issues/41.

Try not swiping if debbugger attached (relaunch app without visual studio then kill then it will work on most devices).

haidarcorp:
I'm sorry. Does this actually fix this issue?? I'm using Android Studio, is it worth plopping this into my project?

guperrot:
@haidarcorp this is a Xamarin PR just to change the implementation of "avoid duplicate push event registration" in a way that does not break the previewer and is very specific to C#.

Android Studio is for Java/Kotlin/C/C++ only and does not have duplicate issues in Java SDK.

PR #641, opened 2018-04-10
Fix capturing exceptions from .NET threads in Android

guperrot:


Apps/Contoso.Android.Puppet/ModulePages/CrashesFragment.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Threading;
```
guperrot: Not available in .net standard 1.x and did not have to migrate to .net standard 2.0 so the fix is tested in the xamarin.android app instead of forms.

PR #642, opened 2018-04-10
[Windows] Fix renewing session id on disable/enable analytics

guperrot:


guperrot:
Need to fix it, it overwrites crash id

PR #643, opened 2018-04-11
[Windows] Refactor ingestion to avoid long tasks

MatkovIvan:
Refined version of #640

PR #644, opened 2018-04-11
[Windows] Fix channel issues

MatkovIvan:
- Do not close current calls from other channels on suspending;
- Do not duplicate `SendingLog`/`FailedToSendLog` events on sending error;
- Add more verbose logs;
- Readable messages on verify event call count in tests;
- Verify if an event occurred more times than expected;

Should be retargeted to develop after merge #643

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Channel/Channel.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Channel/Channel.cs
@@ -16,20 +16,35 @@ namespace Microsoft.AppCenter.Channel
     public sealed class Channel : IChannelUnit
     {
         private const int ClearBatchSize = 100;
+
         private Ingestion.Models.Device _device;
+
         private readonly string _appSecret;
+
         private readonly IStorage _storage;
+
         private readonly IIngestion _ingestion;
+
         private readonly IDeviceInformationHelper _deviceInfoHelper = new DeviceInformationHelper();
+
         private readonly IDictionary<string, List<Log>> _sendingBatches = new Dictionary<string, List<Log>>();
+
         private readonly ISet<IServiceCall> _calls = new HashSet<IServiceCall>();
+
         private readonly int _maxParallelBatches;
+
         private readonly int _maxLogsPerBatch;
+
         private long _pendingLogCount;
+
         private bool _enabled;
+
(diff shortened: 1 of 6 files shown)


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Channel/Channel.cs
```diff
@@ -147,12 +148,16 @@ public void SetEnabled(bool enabled)
                 if (discardLogs)
                 {
                     AppCenterLog.Warn(AppCenterLog.LogTag, "Channel is disabled; logs are discarded");
+                    AppCenterLog.Verbose(AppCenterLog.LogTag, $"Invoke SendingLog event for channel '{Name}'");
```
guperrot: Use debug for that.
Verbose is for http traces or similar use case.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Channel/Channel.cs
```diff
@@ -261,15 +267,17 @@ private void Resume(State state)
...
             }
             CheckPendingLogs(state);
         }
 
         private void Suspend(State state, bool deleteLogs, Exception exception)
         {
+            AppCenterLog.Verbose(AppCenterLog.LogTag, $"Suspend channel: '{Name}'");
             try
             {
+                List<string> sendingBatches = null;
```
guperrot: IList whenever possible

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Channel/Channel.cs
```diff
@@ -287,65 +311,69 @@ private void Suspend(State state, bool deleteLogs, Exception exception)
...
                     {
+                        AppCenterLog.Debug(AppCenterLog.LogTag, $"Invoke FailedToSendLog event for channel '{Name}'");
                         FailedToSendLog?.Invoke(this, new FailedToSendLogEventArgs(log, exception));
                     }
                 }
                 if (deleteLogs)
                 {
-                    try
-                    {
-                        _ingestion.Close();
```
achocron: So we will no longer close ingestion when suspending the channel? This makes sense, but what about shutting it down when the channel group is shut down?

PR #654, opened 2018-04-19
Trying to fix AppCenter Demo build for Xamarin

guperrot:


guperrot:
@dhei please don't delete branch after merging this time. Need to keep after to verify build just in case it fails on master.

PR #656, opened 2018-04-19
[Test Apps] Display null values to avoid completely empty push dialog

MatkovIvan:


Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/App.xaml.cs
```diff
@@ -70,12 +70,12 @@ static void OnPushNotificationReceived(object sender, PushNotificationReceivedEv
         {
             Xamarin.Forms.Device.BeginInvokeOnMainThread(() =>
             {
-                var message = e.Message;
-                if (e.CustomData != null && e.CustomData.Count > 0)
-                {
-                    message += "\nCustom data = {" + string.Join(",", e.CustomData.Select(kv => kv.Key + "=" + kv.Value)) + "}";
-                }
-                Current.MainPage.DisplayAlert(e.Title, message, "OK");
+                var customData = e.CustomData != null
```
guperrot: its ok to skip title I think, since you already have null for message

guperrot: actually its not completely empty since custom data is never null.

guperrot: We should just make sure we display Custom data: {} and skip title and message if null. CustomData is guaranteed not to be null.

guperrot: From original code the only change that i would have done is removing `e.CustomData.Count > 0` from the if and that's it.

PR #659, opened 2018-05-01
Feature/migrate xamarin ios app signing

TroubleMakerBen:


Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.iOS/Contoso.Forms.Demo.iOS.csproj
```diff
@@ -52,11 +53,12 @@
     </DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <CodesignKey>iPhone Developer</CodesignKey>
+    <CodesignKey>iPhone Distribution</CodesignKey>
```
guperrot: We were using a development provisioning profile (even to distribute on appcenter) on purpose to be able to use development push certificate on the same app secret as when we deploy to usb to a device. The name here suggests its a distribution profile being used which would break the use case we had before. Can we have distribute profile but development push? If yes you can leave it like that, if not we should use development profile like before (it works for inapp updates to use development).

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.iOS/Contoso.Forms.Puppet.iOS.csproj
```diff
@@ -37,13 +37,14 @@
     </DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
-    <CodesignKey>iPhone Developer</CodesignKey>
+    <CodesignKey>iPhone Distribution</CodesignKey>
```
guperrot: same comment as demo

Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.iOS/Contoso.Forms.Demo.iOS.csproj
TroubleMakerBen: Just changed this back to debug. No idea why GH doesn't display that change. See https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/659/commits/f2e459c5dfdd63b8107ad9ff5cac4d266cb0207e

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.iOS/Contoso.Forms.Puppet.iOS.csproj
TroubleMakerBen: Same as my previous comment ^^

Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.iOS/Contoso.Forms.Demo.iOS.csproj
guperrot: You changed line 39 in the referenced commit, this is line 56, please check again @TroubleMakerBen.

guperrot: No I am looking at the entire diff. Not just a commit. Please triple check your local files.

PR #660, opened 2018-05-02
Catch all exceptions from (de)serialize exceptions

MatkovIvan:
- Catch all exceptions (there is crash report caused by SDK with `System.Reflection.TargetInvocationException`);
- Do not serialize serialization exception;
- Do not assign deserialization exception as original one;
- Dispose `MemoryStream`s

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared.Targets/CrashesUtils.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared.Targets/CrashesUtils.cs
@@ -8,10 +8,6 @@ namespace Microsoft.AppCenter.Crashes
     {
         internal static byte[] SerializeException(Exception exception)
         {
-            if (exception == null)
-            {
-                return null;
-            }
             if (!exception.GetType().IsSerializable)
             {
                 AppCenterLog.Warn(Crashes.LogTag, $"Cannot serialize {exception.GetType().FullName} exception for client side inspection. " +
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared.Targets/CrashesUtils.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared.Targets/CrashesUtils.cs
@@ -8,6 +8,17 @@ namespace Microsoft.AppCenter.Crashes
     {
         internal static byte[] SerializeException(Exception exception)
         {
+            if (exception == null)
+            {
+                return null;
+            }
+            if (!exception.GetType().IsSerializable)
+            {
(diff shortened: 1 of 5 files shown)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared.Targets/CrashesUtils.cs
```diff
@@ -9,41 +8,37 @@ internal static class CrashesUtils
...
             {
-               deserializedException = formatter.Deserialize(ms) as Exception;
+                using (var memoryStream = new MemoryStream(exceptionBytes))
+                {
+                    var formatter = new BinaryFormatter();
+                    return formatter.Deserialize(memoryStream) as Exception;
+                }
             }
-            catch(SerializationException e)
+            catch (Exception e)
```
guperrot: Do you know what other type is thrown? Should avoid catching knives like outofmemoryexception or similar (thinking about the when(e.IsNoFatal) extension we have in our backend code). For serialize we already crashed so I am not concerned, but deserialize is called in normal app running scenario.

MatkovIvan: Please note, if it's crashed on deserialize we'll not send original crash since it happens on `OnBeforeSending` callback for example. But I agree with you about fatal ones.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared.Targets/CrashesUtils.cs
```diff
@@ -8,6 +8,17 @@ internal static class CrashesUtils
     {
         internal static byte[] SerializeException(Exception exception)
         {
+            if (exception == null)
```
guperrot: How can it be null?

MatkovIvan: With current usage - no way. This is just a guard

MatkovIvan: `as` operator is used [here](https://github.com/Microsoft/AppCenter-SDK-DotNet/blob/724d830d80103939f236926a0e8e4a6069d4672b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.iOS/Crashes.cs#L102). So, it shouldn't be `null` but I would prefer that this check remains

guperrot: That does not help if it ever goes null as `GenerateiOSException` does not check for null. If we have any reason to think it might actually be null, add check there too.

PR #665, opened 2018-05-10
Fix a crash when Crashes.TrackError exception parameter is null

guperrot:
Print an error log instead that the parameter is not allowed to be null.

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared/Crashes.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared/Crashes.cs
@@ -7,172 +7,179 @@ using System.Threading.Tasks;
 
 namespace Microsoft.AppCenter.Crashes
 {
-    /// <summary>
-    /// Crashes service.
-    /// </summary>
-    public partial class Crashes
-    {
-        /// <summary>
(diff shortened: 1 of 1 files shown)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared/Crashes.cs
```diff
@@ -7,172 +7,179 @@
...
+		/// <param name="properties">Optional properties.</param>
 #endif
-        public static void TrackError(Exception exception, IDictionary<string, string> properties = null)
-        {
-            PlatformTrackError(exception, properties);
-        }
-    }
+		public static void TrackError(Exception exception, IDictionary<string, string> properties = null)
+		{
+			if (exception == null)
```
guperrot: File was formatted incorrectly :/
This is the method that was actually patched with the if and log message.

dhei: Looks like indentation is 6.

guperrot: It's correct this time with ctrl+I.

PR #693, opened 2018-06-20
Update demo application to Xamarin.Forms 3.0

MatkovIvan:
- Update to Xamarin.Forms 3.0;
- Update portable project to .NETStandard 2.0;
- Replace `project.json`s to `PackageReference`s;


Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.UWP/Contoso.Forms.Demo.UWP.csproj
```diff
@@ -11,13 +11,14 @@
     <AssemblyName>Contoso.Forms.Demo.UWP</AssemblyName>
     <DefaultLanguage>en-US</DefaultLanguage>
     <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
-    <TargetPlatformVersion>10.0.14393.0</TargetPlatformVersion>
-    <TargetPlatformMinVersion>10.0.10586.0</TargetPlatformMinVersion>
+    <TargetPlatformVersion>10.0.16299.0</TargetPlatformVersion>
+    <TargetPlatformMinVersion>10.0.16299.0</TargetPlatformMinVersion>
```
guperrot: Any specific reason to bump minimum version? Creators Fall is still recent, we have older phones.

AppCenter-Demo.sln
```diff
@@ -7,7 +7,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Apps", "Apps", "{274AB35C-6
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Contoso.Forms.Demo", "Contoso.Forms.Demo", "{91CEE2C9-F64C-45C9-9353-B1EC87D13C6D}"
 EndProject
```
guperrot: No change needed in the Mac or Windows solution?

Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.UWP/Contoso.Forms.Demo.UWP.csproj
MatkovIvan: .NET Standard 2.0 on UWP requires [the minimum version to be set to Fall Creator's Update](https://blogs.msdn.microsoft.com/dotnet/2017/10/10/announcing-uwp-support-for-net-standard-2-0/). Also Xamarin.Forms 3.0 on UWP [has a dependency on .NET Standard 2.0](https://github.com/xamarin/Xamarin.Forms/issues/2859).

I guess this is OK for *one* of our test apps.

guperrot: We don't have a non forms app that uses nuget references, the one we have is project references only. I think it would be much safer to have another blank app (just a hardcoded button in center to track a predefined event, and push/crashes enabled with no customization) that we can use to test compatibility with the arm devices we have on older versions we support.

Apps/Contoso.UWP.Demo/App.xaml.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// Initializes the singleton application object.  This is the first line of authored code
+        /// executed, and as such is the logical equivalent of main() or WinMain().
+        /// </summary>
+        public App()
+        {
+            this.InitializeComponent();
+            this.Suspending += OnSuspending;
+            AppCenter.LogLevel = LogLevel.Verbose;
+            AppCenter.SetLogUrl("https://in-integration.dev.avalanch.es");
+            AppCenter.Start("42f4a839-c54c-44da-8072-a2f2a61751b2", typeof(Analytics), typeof(Crashes), typeof(Push));
```
guperrot: This app has already a store association with the other puppet, and store association matters for Push so we cannot reuse the same app.

I would create another one, also since we use nugets by convention we use appcenter.ms for those apps.

Apps/Contoso.UWP.Demo/Contoso.UWP.Demo.csproj
```diff
@@ -0,0 +1,146 @@
...
+    <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
+    <ProjectGuid>{612D30FE-E1F1-4CEE-8256-911C03A16890}</ProjectGuid>
+    <OutputType>AppContainerExe</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Contoso.UWP.Demo</RootNamespace>
+    <AssemblyName>Contoso.UWP.Demo</AssemblyName>
+    <DefaultLanguage>en-US</DefaultLanguage>
+    <TargetPlatformIdentifier>UAP</TargetPlatformIdentifier>
+    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">10.0.14393.0</TargetPlatformVersion>
+    <TargetPlatformMinVersion>10.0.10586.0</TargetPlatformMinVersion>
```
guperrot: Our docs document a lower version, we should match that here.

Apps/Contoso.UWP.Demo/Contoso.UWP.Demo.csproj
```diff
@@ -0,0 +1,146 @@
...
+    <Compile Include="MainPage.xaml.cs">
+      <DependentUpon>MainPage.xaml</DependentUpon>
+    </Compile>
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <AppxManifest Include="Package.appxmanifest">
+      <SubType>Designer</SubType>
+    </AppxManifest>
+    <None Include="Contoso.UWP.Demo_TemporaryKey.pfx" />
```
guperrot: Please remove this file include for now like we did in other apps.

Apps/Contoso.UWP.Demo/Contoso.UWP.Demo.csproj
MatkovIvan: This version was set in the XF UWP Demo project
https://github.com/Microsoft/AppCenter-SDK-DotNet/blob/f3aba9f025a99cc1ead295beb1f9ad9b53c38677/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.UWP/Contoso.Forms.Demo.UWP.csproj#L15

Anyway, I'll downgrade

PR #694, opened 2018-06-30
Firebase dependency is now mandatory again

guperrot:


guperrot:
@jaelim-ms it's a reduced version with no secrets. This will make contributions much easier.

scripts/configuration/ac-build-config.xml
```diff
@@ -3,7 +3,7 @@
     <sdkInfo>
         <sdkVersion>1.7.1-SNAPSHOT</sdkVersion>
         <iosVersion>1.7.1</iosVersion>
-        <androidVersion>1.6.1</androidVersion>
+        <androidVersion>1.7.0-5+902d4065</androidVersion>
```
guperrot: Need to switch to 1.7.0 after official release before merging

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet.Droid/MainActivity.cs
```diff
@@ -30,7 +30,6 @@ protected override void OnCreate(Bundle savedInstanceState)
             Xamarin.Forms.Forms.Init(this, savedInstanceState);
 
             AndroidAnalytics.SetListener(new AndroidAnalyticsListener());
-            Push.SetSenderId("177539951155");
```
TroubleMakerBen: We didn't remove this in Sasquatch Android for backwards compat, we probably want to do this here, too.

guperrot: @TroubleMakerBen Sasquatch has 2 build variants, we don't have that in Xamarin.

SDK/AppCenterPush/Microsoft.AppCenter.Push.Android/Push.cs
```diff
@@ -53,6 +53,7 @@ static Task PlatformSetEnabledAsync(bool enabled)
         /// <value>
         /// The Sender ID to set.
         /// </value>
+        [Obsolete("For all the Android developers using App Center, there is a change coming where Firebase SDK is required to use Push Notifications. For Android P, its scheduled at the release date for the latest OS version. For all other versions of Android, it will be required after April 2019. Please follow the migration guide at https://aka.ms/acfbxa.")]
```
jwallra: This could be more concise.  I would suggest "There is a change in Android P on release which requires the Firebase SDK to be used for Push notifications.  For other versions, the change will go into effect in April 2019.  Please follow the migration guide at https://aka.ms/acfbxa."

jwallra: As-is, in second sentence, its->it's

PR #700, opened 2018-07-12
Catch exception when checking network connection

TroubleMakerBen:


TroubleMakerBen:
Tagged as `dont-merge-yet` as I'm not sure how to best unit-tests handling the excpetion. I'll grab someone from our team tomorrow but feel free to leave a comment =)

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Ingestion/Http/NetworkStateAdapter.cs
```diff
@@ -9,7 +9,23 @@ public NetworkStateAdapter()
...
+        {
+            get
+            {
+                try
+                {
+                    return NetworkInterface.GetIsNetworkAvailable();
+                }
+                catch (Exception e)
+                {
+                    AppCenterLog.Error(AppCenterLog.LogTag, "An error occurred while updating network state.", e);
```
guperrot: Isn't that "checking" instead of "updating"?

guperrot:
@TroubleMakerBen As a unit test, we need to be able to mock the API throwing the exception and then checking we get a false from this method. Unlike Android, static methods cannot be mocked so we might need an indirection, could be an internal delegate property that defaults to pointing to the real method and we replace that in the unit test by a mock. @achocron has more context on that.

guperrot:
We should merge it now otherwise it will miss the release train, the unit test with mock network is not critical at this point.

PR #711, opened 2018-07-27
Fix/write location for windows

achocron:


SDK/AppCenter/Microsoft.AppCenter.UWP/Utils/ApplicationStoragePathHelper.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.AppCenter.Utils
+{
+    public class ApplicationStoragePathHelper
+    {
+        public static string ApplicationStoragePath => "";
```
jwallra: Why create a storage path helper which just has an empty path constant?

achocron: It needs to have the same interface as the other platforms since we use shared code

dhei: nit: `string.Empty` instead of `""`

Tests/Microsoft.AppCenter.NET/ApplicationStoragePathHelper.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.AppCenter.Utils
+{
+    public class ApplicationStoragePathHelper
+    {
+        public static string ApplicationStoragePath => "";
```
dhei: nit - `string.Empty`

PR #715, opened 2018-08-06
Stabilize UITests in App Center

owenniblock:
### Motivation
The UITests are failing intermittently - this work attempts to stabilise the tests.

### Notes
I'm having trouble getting the `TestServiceStatePersistence` tests to go green on all Android devices. I suspect this might be an issue with the Calabash Android Server (I've created a bug on our side for investigation).

iOS tests are consistently green on a local device and in App Center.

Android tests (except the above) are consistently green on a local device and in App Center.

If you let me know your App Center user name - I'll add you as a collaborator so you can see the app test runs and the device selection. It's possible that I've been lucky with my device selection but it seems to cover a relatively broad cross-section.

Also - I have flagged up for investigation an issue where the Toggle page seems to continue before the data is saved (and then when the app restarts it fails to complete the save). I've found that if I add a long wait in then it works correctly so maybe an async process isn't waiting for completion correctly (definitely worth checking this out). For now the long wait will keep the tests green unless there's additional issues.

### DIFF
--- a/Tests/Contoso.Forms.Test/Module Pages/CrashesPage.xaml.cs
+++ b/Tests/Contoso.Forms.Test/Module Pages/CrashesPage.xaml.cs
@@ -9,6 +9,8 @@ namespace Contoso.Forms.Test
 {
     public partial class CrashesPage : ContentPage
     {
+        private const int BeforeCrashSleepTime = 500;
+
         public CrashesPage()
         {
             InitializeComponent();
@@ -21,7 +23,7 @@ namespace Contoso.Forms.Test
 
         void DivideByZeroCrash(object sender, System.EventArgs e)
         {
-            Task.Delay(500).Wait();
+            Task.Delay(BeforeCrashSleepTime).Wait();
 
 #pragma warning disable CS0219
             int x = (42 / int.Parse("0"));
@@ -35,7 +37,7 @@ namespace Contoso.Forms.Test
 
         void CrashWithInvalidOperation(object sender, EventArgs e)
         {
-            Task.Delay(500).Wait();
+            Task.Delay(BeforeCrashSleepTime).Wait();
 
             string[] strings = { "A", "B", "C" };
 #pragma warning disable CS0219
@@ -57,7 +59,7 @@ namespace Contoso.Forms.Test
 
         private void CrashWithAggregateException(object sender, EventArgs e)
         {
-            Task.Delay(500).Wait();
+            Task.Delay(BeforeCrashSleepTime).Wait();
 
             throw PrepareException();
         }
@@ -100,7 +102,7 @@ namespace Contoso.Forms.Test
 
         public async void CrashAsync(object sender, EventArgs e)
         {
-            Task.Delay(500).Wait();
+            Task.Delay(BeforeCrashSleepTime).Wait();
 
             await FakeService.DoStuffInBackground().ConfigureAwait(false);
         }
--- a/Tests/Contoso.Forms.Test/Module Pages/CrashesPage.xaml.cs
+++ b/Tests/Contoso.Forms.Test/Module Pages/CrashesPage.xaml.cs
@@ -21,6 +21,8 @@ namespace Contoso.Forms.Test
 
         void DivideByZeroCrash(object sender, System.EventArgs e)
         {
+            Task.Delay(500).Wait();
+
 #pragma warning disable CS0219
             int x = (42 / int.Parse("0"));
 #pragma warning restore CS0219
@@ -33,6 +35,8 @@ namespace Contoso.Forms.Test
 
         void CrashWithInvalidOperation(object sender, EventArgs e)
         {
+            Task.Delay(500).Wait();
+
             string[] strings = { "A", "B", "C" };
 #pragma warning disable CS0219
             string s = strings.First((arg) => { return arg == "6"; });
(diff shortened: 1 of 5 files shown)


Tests/UITests/Tests.cs
```diff
@@ -68,54 +70,95 @@ public void TestEnablingAndDisablingServices()
...
             app.Tap(TestStrings.GoToTogglePageButton);
 
             /* Make sure Crashes enabled state is persistent */
             ServiceStateHelper.AppCenterEnabled = true;
             ServiceStateHelper.CrashesEnabled = false;
             Assert.IsFalse(ServiceStateHelper.CrashesEnabled);
+            //There seems to be some sort of timing issue here.
+            //   without the Thread.Sleep - this code fails although I see that the method called is Async (with a Wait)
+            //   so I think this deserves further investigation. I've added a terribly long wait to get the tests green.
+            Thread.Sleep(10000);
```
jwallra: I'm also skeptical that the thread need to be slept for this long.  If that's the case then there may be a different issue that we need to address.  Also I would like to align the sleep times with the Exception tests.

Tests/Contoso.Forms.Test/Module Pages/CrashesPage.xaml.cs
```diff
@@ -21,6 +21,8 @@ void DismissPage(object sender, System.EventArgs e)
 
         void DivideByZeroCrash(object sender, System.EventArgs e)
         {
+            Task.Delay(500).Wait();
```
achocron: Should  these numbers be in a constant or something?

Tests/UITests/Tests.cs
achocron: @jwallra It is long but I'd rather our UITests be slow than fail intermittently since they are running over night anyways

Tests/UITests/Tests.cs
```diff
@@ -68,54 +70,95 @@ public void TestEnablingAndDisablingServices()
...
+            app.Screenshot("App Launched - Ready for tests");
             ServiceStateHelper.app = app;
+            app.WaitForElement(TestStrings.GoToTogglePageButton);
             app.Tap(TestStrings.GoToTogglePageButton);
 
             /* Make sure Crashes enabled state is persistent */
             ServiceStateHelper.AppCenterEnabled = true;
             ServiceStateHelper.CrashesEnabled = false;
             Assert.IsFalse(ServiceStateHelper.CrashesEnabled);
+            //There seems to be some sort of timing issue here.
```
jwallra: Please add a blank line before this comment and format with a single space after each `//`

Tests/UITests/Tests.cs
```diff
@@ -68,54 +70,95 @@ public void TestEnablingAndDisablingServices()
...
+            app.Screenshot("App Launched - Ready for tests");
+            ServiceStateHelper.app = app;
+            app.WaitForElement(TestStrings.GoToTogglePageButton);
+            app.Tap(TestStrings.GoToTogglePageButton);
 
             /* Make sure Analytics enabled state is persistent */
             ServiceStateHelper.AppCenterEnabled = true;
             ServiceStateHelper.AnalyticsEnabled = false;
             Assert.IsFalse(ServiceStateHelper.AnalyticsEnabled);
+            Thread.Sleep(10000);
```
jwallra: Please move these sleep values to constants in this file.  Align if possible or name appropriately.

guperrot:
@achocron @jwallra lets run or change nightly run to https://github.com/Microsoft/AppCenter-SDK-DotNet/tree/owenniblock-fix/stabilize-uitests before merging to verify an actual test cloud run.

Tests/UITests/Tests.cs
owenniblock: That was my thinking too - I can tinker with the times but I wanted to be sure that it cleared the issue (network speeds can be slow so if it's something to do with that it's difficult to say how long we need to wait for here).

guperrot:
The PR does not change anything in test cloud.
We have a known issue with UI tests and NSDefault persistence on iOS 10+ and I don't think it's a delay problem, those persistence tests still fail with this branch: https://appcenter.ms/users/alchocro/apps/mobilecenter-xamarin-testing-app-ios/test/runs/5a3e144a-58c8-4d06-9b3f-54a52d905149

Overall we have not seen any improvement based on the test suite result.

owenniblock:
@guperrot - I did not hit that issue on iOS myself :( Let me take a look at the back-end logs and see if I can work out why!

owenniblock:
Are you sure that's the correct test run? I've replaced `Test Service State Persistence` with 3 separate tests to debug each bit (while debugging this issue on Android).

guperrot:
@owenniblock launching a new one.

guperrot:
@owenniblock last run passed for both platforms so I guess we can merge soon. Please address the feedbacks from Jacob and the constant one from Zander and then we can merge.

owenniblock:
Cool, I'm still chasing down one last issue on Android (partly for my own piece of mind). Give me a few days and hopefully then we'll be ready to merge!

guperrot:
Our nightly run has been switched to your branch for now.

owenniblock:
I think I'm done. If you add me as a collaborator on the app in App Center you're using for tests I'll be able to keep an eye on it for any more subtle instability issues over the next few weeks.

achocron:
@owenniblock thanks for the contribution!

PR #730, opened 2018-08-22
[Test app] Catch exceptions on adding event properties

MatkovIvan:


Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/AnalyticsContentPage.xaml.cs
```diff
@@ -51,7 +52,14 @@ void TrackEvent(object sender, EventArgs e)
             var properties = new Dictionary<string, string>();
             foreach (Property property in EventProperties)
             {
-                properties.Add(property.Name, property.Value);
+                try
+                {
+                    properties.Add(property.Name, property.Value);
+                }
+                catch (Exception ex)
```
guperrot: Can we use a if instead of catching everything?

MatkovIvan: That would not just one `if` then. at least it needs for null key check and check for already contains key. The current way gives enough good error description, so I don't see anything wrong with catch exceptions from just one line with enough simply operation.

MatkovIvan:
`.xaml.cs` is default template. This rename harmonize it with other pages and fixes folding (.xaml is parent for .xaml.cs in the project tree-view)

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/AnalyticsContentPage.xaml.cs
guperrot: This looks weird to have to log an exceptional case while it's a standard UI action triggering both the conditions.

Plus one could argue  if (key != null && !dict.ContainsKey) uses 3 less lines of code as no catch block and no logging code.

MatkovIvan: OK. But then I prefer to move both checks to adding the property rather than sending

guperrot: Sure

PR #732, opened 2018-08-23
Bump sdk version to 1.9.0

MatkovIvan:

PR #733, opened 2018-08-23
Merge develop -> master

MatkovIvan:
Merge only after #732


guperrot:
Experiencing issues with builds, optional checks should be green too for release.

PR #736, opened 2018-08-27
Remove tests and forms demo applciation from windows solution

MatkovIvan:


achocron:
@MatkovIvan Is this because there is also an iOS and Android project in the XF Demo? If so, is it possible to just exclude those?

PR #750, opened 2018-09-06
Catch exceptions from GetLocaleInfoEx

evgeny-pol:
Do not fail attempt to send analytics events if `api-ms-win-core-localization-l1-2-0.dll` library is missing or failed to load

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
@@ -46,7 +46,7 @@ namespace Microsoft.AppCenter.Windows.Shared.Utils
                     return buffer.ToString();
                 }
             }
-            catch (Exception exception)
+            catch (DllNotFoundException exception)
             {
                 AppCenterLog.Verbose(AppCenterLog.LogTag, $"Failed to call GetLocaleInfoEx: {exception.Message}");
             }
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
@@ -1,4 +1,5 @@
-﻿using System.Globalization;
+﻿using System;
+using System.Globalization;
 using System.Runtime.InteropServices;
 using System.Text;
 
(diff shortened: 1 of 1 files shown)


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -36,11 +37,18 @@ public static CultureInfo GetCurrentCulture()
...
+            try
+            {
+                var buffer = new StringBuilder(BUFFER_SIZE);
+                var resultCode = GetLocaleInfoEx(lpLocaleName, LCType, buffer, BUFFER_SIZE);
+                if (resultCode > 0)
+                {
+                    return buffer.ToString();
+                }
+            }
+            catch (Exception exception)
```
MatkovIvan: Please catch more specific type

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -36,11 +37,18 @@ public static CultureInfo GetCurrentCulture()
...
+                var resultCode = GetLocaleInfoEx(lpLocaleName, LCType, buffer, BUFFER_SIZE);
+                if (resultCode > 0)
+                {
+                    return buffer.ToString();
+                }
+            }
+            catch (DllNotFoundException exception)
             {
-                return buffer.ToString();
+                AppCenterLog.Verbose(AppCenterLog.LogTag, $"Failed to call GetLocaleInfoEx: {exception.Message}");
```
MatkovIvan: I guess the log level should be `Debug` or `Warning`. @guperrot Which one do you prefer?

guperrot: It should be a warning yes.

guperrot: After talking to Ivan offline, we can leave it as debug if that always happens when bound to Unity.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/CultureInfoHelper.cs
```diff
@@ -15,7 +16,7 @@ class CultureInfoHelper
         private const string LOCALE_NAME_SYSTEM_DEFAULT = "!x-sys-default-locale";
 
         private const int BUFFER_SIZE = 530;
-        
+
         [DllImport("api-ms-win-core-localization-l1-2-0.dll", CharSet = CharSet.Unicode)]
         private static extern int GetLocaleInfoEx(string lpLocaleName, uint LCType, StringBuilder lpLCData, int cchData);
```
guperrot: Lines 31/32: please update code and comment to

```
                    // If system/user default doesn't work, use current culture which is app locale.
                    return CultureInfo.CurrentCulture;
```

To avoid sending empty string as fallback.

Issue #753, opened 2018-09-18
[UWP] No compatible with sqlite 1.5.231


MarMarIV:
### **Description**

After upgrade sqlite and xamarin.forms to latest version, method AppCenter.Start throw exception:

Method not found: 'Void SQLite.SQLiteAsyncConnection..ctor(System.String, Boolean)

### **Repro Steps**

1. NuGet Xamarin Forms 3.2
2. NuGet Sqlite-net-pcl 1.5.231
3. NuGet AppCenter 1.9
4. override OnStart method in App.xaml.cs:

AppCenter.Start(
                   "uwp=...;" +
                   "android=...;" +
                   "ios=...;",
                   typeof(Analytics), typeof(Crashes));

### **Details**

With sqlite-net-pcl 1.5.166-beta it is working

dhei:
Hey @MarMarIV,

Thanks for reporting this issue. Looks like sqlite-net-pcl v1.5.231 has some breaking change in the `SQLiteAsyncConnection` [constructor](SQLiteAsyncConnection) that breaks App Center SDK. (See https://github.com/praeclarum/sqlite-net/issues/751 for users reporting this to sqlite-net-pcl repo).

We are working on a fix and we will share more info when we have the fix. Thanks for the understanding.

MatkovIvan:
Hey!
We've included fixes in the latest SDK release for this. I'm closing the issue but please reopen if you run into this again.

MatkovIvan:
Hi @gmarbury,

The `sqlite-net-pcl` is only used for UWP and does not affect Xamarin Live Player (Android) in any way.

guperrot:
The android and iOS dlls are mostly implemented using bindings to the Java and Objective C SDKs which is considered native code by Xamarin Live Player.

PR #754, opened 2018-09-19
Fix compatibility with SQLite 1.5.231

MatkovIvan:
#753

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
@@ -13,7 +13,7 @@ namespace Microsoft.AppCenter.Storage
 
         public StorageAdapter(string databasePath)
         {
-            // In SQLite-net 1.5.231 constructor parameters was changed.
+            // In SQLite-net 1.5.231 constructor parameters were changed.
             // Using reflection to accept newer library version.
             _dbConnection = (SQLiteAsyncConnection)typeof(SQLiteAsyncConnection)
                 .GetConstructor(new[] { typeof(string), typeof(bool) })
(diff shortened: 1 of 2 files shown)


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -13,7 +13,21 @@ internal class StorageAdapter : IStorageAdapter
 
         public StorageAdapter(string databasePath)
         {
-            _dbConnection = new SQLiteAsyncConnection(databasePath);
+            // In SQLite-net 1.5.231 constructor parameters was changed.
+            // Using reflection to accept newer library version.
+            _dbConnection = (SQLiteAsyncConnection)typeof(SQLiteAsyncConnection)
+                .GetConstructor(new[] { typeof(string), typeof(bool) })
+                ?.Invoke(new object[] { databasePath, true });
```
guperrot: I don't see boolean parameter in https://github.com/oysteinkrog/SQLite.Net-PCL/blob/master/src/SQLite.Net.Async/SQLiteAsyncConnection.cs.

Previous version of the code has single parameter, in your patch you either pass 2 or 3 parameters (so more than 1 even in old version), I am confused about that, can you comment more about what code is for what version and what are the parameters for?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -13,7 +13,21 @@ internal class StorageAdapter : IStorageAdapter
 
         public StorageAdapter(string databasePath)
         {
-            _dbConnection = new SQLiteAsyncConnection(databasePath);
+            // In SQLite-net 1.5.231 constructor parameters was changed.
```
guperrot: was => were

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
MatkovIvan: You linked some fork where last update was in 2015.
The second parameter had default value. See https://github.com/praeclarum/sqlite-net/commit/c360ae602dd3a36d8ffdb60e3101c872a5de872a#diff-2388537c7df63c2e1c2c63b815f301b4L57

PR #756, opened 2018-10-02
Add push to ios puppet (non forms)

guperrot:


Apps/Contoso.iOS.Puppet/AppDelegate.cs
```diff
@@ -1,86 +1,95 @@
...
+            if (UIDevice.CurrentDevice.CheckSystemVersion(10, 0))
+            {
+                UNUserNotificationCenter.Current.Delegate = new PushDelegate();
+            }
+            Push.PushNotificationReceived += (sender, e) =>
+            {
+                string message = "";
+                if (UIDevice.CurrentDevice.CheckSystemVersion(10, 0))
+                {
+                    if (_didTapNotification)
```
MatkovIvan: The ternary operator can improve readability here.
`message += _didTapNotification ? "Tapped notification\n" : "Received in foreground\n";`
This also removes `message.Length > 0` below

guperrot: Will use ternary but on iOS 9 the second suggestion will add an extra new line so it's slightly different.

guperrot: If message is null but have custom data we end up having 1 extra newline, this needs a bit more checking

PR #799, opened 2019-01-10
Don't start session from logs

MatkovIvan:

## Description

1. Schedule track event every 30s
2. Minimize the app into the background
3. Observe that every event creates a new session

Expected behavior:
Like iOS, sending a log just extends current session lifetime but must not trigger a new session.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### UWP
 
-* **[Fix]** Extend the current session instead of starting a new session when sending events from the background. Sessions are also no longer started in background by sending an event or a log from another service such as push, as a consequence the push registration information will be missing from crash events information.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 
 #### UWP
 
-* **[Fix]** Don't start a new session from logs in the background.
+* **[Fix]** Extend the current session instead of starting a new session when sending events from the background. Sessions are also no longer started in background by sending an event or a log from another service such as push, as a consequence the push registration information will be missing from crash events information.
```
guperrot: We don't have crash breadcrumbs for windows platforms. The message might be relevant in the future but for now the second part is confusing for UWP.

PR #833, opened 2019-02-12
Updated Native SDK 1.14.0 updated

Subhashvysyaraju:
## Description

Updated Native SDK 1.14.0 updated


guperrot:
This version does not contain any new features, so 1.13.1 was already the right version.

PR #841, opened 2019-02-21
Update ac-build-config.xml

Subhashvysyaraju:
## Description

Updated IOS Version 1.13.2

### DIFF
--- a/scripts/configuration/ac-build-config.xml
+++ b/scripts/configuration/ac-build-config.xml
@@ -3,7 +3,7 @@
     <sdkInfo>
         <sdkVersion>1.13.2-SNAPSHOT</sdkVersion>
         <iosVersion>1.13.2</iosVersion>
-        <androidVersion>1.11.2</androidVersion>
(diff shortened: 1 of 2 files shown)


scripts/configuration/ac-build-config.xml
```diff
@@ -2,7 +2,7 @@
 <config>
     <sdkInfo>
         <sdkVersion>1.13.2-SNAPSHOT</sdkVersion>
-        <iosVersion>1.13.0</iosVersion>
+        <iosVersion>1.13.2</iosVersion>
         <androidVersion>1.11.2</androidVersion>
```
guperrot: https://github.com/Microsoft/AppCenter-SDK-Android/releases => we didn't bump to 1.11.3 last time, was that missed on first release attempt?

Subhashvysyaraju: Yes

guperrot: Good thing we see it now then.

PR #848, opened 2019-02-28
Fix bug 57688 : [DotNet] User Id doesn't work for SDK 1.13.2

JasoWu:

## Description

We found that press enter after input any key the key action will be 'DOWN', Otherwise it will be 'UP',
So add another choice to set the user id.
image

### DIFF
--- a/Apps/Contoso.Android.Puppet/ModulePages/AppCenterFragment.cs
+++ b/Apps/Contoso.Android.Puppet/ModulePages/AppCenterFragment.cs
@@ -114,7 +114,7 @@ namespace Contoso.Android.Puppet
 
         private void UserIdTextKeyPressedHandler(object sender, View.KeyEventArgs e)
         {
-            if (e.Event.Action == KeyEventActions.Down && e.KeyCode == Keycode.Enter)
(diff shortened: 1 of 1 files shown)


Apps/Contoso.Android.Puppet/ModulePages/AppCenterFragment.cs
```diff
@@ -114,7 +114,7 @@ private void LogWriteLevelClicked(object sender, EventArgs e)
 
         private void UserIdTextKeyPressedHandler(object sender, View.KeyEventArgs e)
         {
-            if (e.Event.Action == KeyEventActions.Down && e.KeyCode == Keycode.Enter)
+            if ((e.Event.Action == KeyEventActions.Down && e.KeyCode == Keycode.Enter) || (e.Event.Action == KeyEventActions.Up))
```
MatkovIvan: The first condition became redundant now

JasoWu: Yeah, You're right, Thank you!

JasoWu:
> What about all the other applications? I don't see an update on demo app while the bug was reported against the demo. You modified only 1 of the puppets.

JasoWu:
@guperrot I have already found the problems for other platform applications, And will send PR today. This PR is just to get your suggestions.

PR #850, opened 2019-03-04
Fix setting user id in Contoso.Forms.Demo

JasoWu:

## Description

Contoso.iOS.Puppet works fine for setting user id.

Removed the third party control.
image

### DIFF
--- a/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/ModulePages/AppCenterContentPage.xaml.cs
+++ b/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/ModulePages/AppCenterContentPage.xaml.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.ComponentModel;
 using Microsoft.AppCenter;
 using Xamarin.Forms;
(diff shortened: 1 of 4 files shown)


Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/ModulePages/AppCenterContentPage.xaml.cs
```diff
@@ -7,13 +7,20 @@ namespace Contoso.Forms.Demo
     [Android.Runtime.Preserve(AllMembers = true)]
     public partial class AppCenterContentPage : ContentPage
     {
+
+        Fabulous.CustomControls.CustomEntryCell customEntryCell;
```
guperrot: Please do not use new third party libs just for that.
Plus it does not build:

screen shot 2019-03-04 at 12 09 54 pm

JasoWu: Removed

PR #862, opened 2019-03-26
fix No intellisense for Crash, Distribute, Push

Zongrenfeng:
Things to consider before you submit the PR:

* [x] Has `CHANGELOG.md` been updated?
* [ ] Are tests passing locally?
* [x] Are the files formatted correctly?
* [ ] Did you add unit tests if this modifies the UWP implementation?
* [x] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description
The previous missed add reference to the xml files inside nuspec files of crash, distribute, push. Also 3 project files need to be update too.

## Related PRs or issues
https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/827/

PR #865, opened 2019-04-01
fix crashes.xml not found error in release

Zongrenfeng:

* [X] Are tests passing locally?
* [X] Are the files formatted correctly?

## Description
The previous fix passed in build, but release process meet an cannot find Microsoft.AppCenter.Crashes.xml file issue, which I found it supposed to be generated by a target "ref/uap10.0" not exist in corresponding project file. So remove this one since does not seems the xml will be generated.

## Related PRs or issues

### DIFF
--- a/nuget/AppCenterCrashes.nuspec
+++ b/nuget/AppCenterCrashes.nuspec
@@ -45,7 +45,6 @@
     <!-- targets files add correct references to proper platforms -->
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.targets" target="build/uap10.0"/>
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.dll" target="ref/uap10.0" />
-    <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.xml" target="ref/uap10.0" />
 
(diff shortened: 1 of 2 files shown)


MatkovIvan:
Please update `WindowsAppCenterCrashes.nuspec` too

Zongrenfeng:
> Please update WindowsAppCenterCrashes.nuspec too
> Please consider fixing Microsoft.AppCenter.Crashes.UWP.csproj instead. It contains a lot of DocumentationFile tags (I guess the issue is related to platform-dependent compilation)

There are 8 targets ([Debug|Release] * [AnyCPU|ARM|x86|x64]) exists inside the UWP project file -- I had add documentationFile tag for each of them, and looks reasonable-- and there is *none* of them match the one (`ref/uap10.0`) which i removed from .nuspec file in this commit. I think that unless we add a new taget into the UWP project, otherwise this one in nuspec will always lead to nuget package fail.

guperrot:
We need to make sure we can see intellisense on a UWP project that does not use portable code. We can merge that one and test, but if that does not work the task will need to remain open. @MatkovIvan thoughts?

PR #867, opened 2019-04-04
fix the crashes module does not show intellisense

Zongrenfeng:

## Description
Observed the Crashes module still does not show intellisense in Android and iOS apps. 

Tried build on Mac machine, then everything is ok. But use the assemblies from remote feed still got this issue.

The root cause is that there are three nuspec files exists for Crashes:
- AppCenterCrashes.nuspec
- WindowsAppCenterCrashes.nuspec
- MacAppCenterCrashes.nuspec

As Ivan said, the windows one contain UWP targets which only can be build in windows, and Mac one only can be build on macOS. CI build them separately, then merge. Then the AppCenterCrashes.nuspec will be used as final nuget package's spec.

Currently, we have added the *DocumentationFile* tag for all targets inside the *Mac version's nuspec*, but did not add the *DocumentationFile* tag for Android and iOS target in the final one `AppCenterCrashes.nuspec`. So if this file used as a `Copy files list` for build script, then the next nuget pack cli will miss the xml files. Add tags into the final nuspec should fix this issue. 

## Related PRs or issues
https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/865

### DIFF
--- a/nuget/AppCenterCrashes.nuspec
+++ b/nuget/AppCenterCrashes.nuspec
@@ -31,10 +31,12 @@
 
     <!-- Android -->
     <file src="$android_dir$/Microsoft.AppCenter.Crashes.dll" target="lib/MonoAndroid403" />
+    <file src="$android_dir$/Microsoft.AppCenter.Crashes.xml" target="lib/MonoAndroid403" />
     <file src="$android_dir$/Microsoft.AppCenter.Crashes.Android.Bindings.dll" target="lib/MonoAndroid403" />
 
     <!-- iOS -->
     <file src="$ios_dir$/Microsoft.AppCenter.Crashes.dll" target="lib/Xamarin.iOS10" />
+    <file src="$ios_dir$/Microsoft.AppCenter.Crashes.xml" target="lib/Xamarin.iOS10" />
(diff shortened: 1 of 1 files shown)


jwallra:
@Zongrenfeng So it sounds like this still doesn't fix the issue.  Could you add the tags to the nuspecs so this PR is actually a fix?  You can use the NuGet package explorer which you can find here: https://github.com/NuGetPackageExplorer/NuGetPackageExplorer

Zongrenfeng:
> @Zongrenfeng So it sounds like this still doesn't fix the issue, is that correct?  Could you add the tags to the nuspecs so this PR is actually a fix?  You can use the NuGet package explorer which you can find here: https://github.com/NuGetPackageExplorer/NuGetPackageExplorer

Oh, maybe I descripted in miss understanding way... What I means is that ideally it's not needed to add the reference to xml file into the final nuspec file because we already have that in the Mac version's nuspec file, and in the merge process on CI server, the build script should will copy the xmls side-by-side with the Android/iOS version dlls to the packge.

Unfortunately, this is not true, the xmls missed from the nuget packages on feeds. So I had to add reference into the final nuspec file to force the build script do copy the xmls. That's what I did in this PR.

Thanks~

PR #876, opened 2019-04-10
fix intellisense for crashes uap10

Zongrenfeng:

## Description
The last intellisense issue for crashes module in target uap10 has an issue on CI server that cannot find the xml file at the same place as the dll, so use the same one from x86 child folder as confirmed locally.

## Related PRs or issues
https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/865
https://github.com/Microsoft/AppCenter-SDK-DotNet/pull/867

### DIFF
--- a/nuget/AppCenterCrashes.nuspec
+++ b/nuget/AppCenterCrashes.nuspec
@@ -47,6 +47,7 @@
     <!-- targets files add correct references to proper platforms -->
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.targets" target="build/uap10.0"/>
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.dll" target="ref/uap10.0" />
+    <file src="$uwp_dir$/x86/Microsoft.AppCenter.Crashes.xml" target="ref/uap10.0" />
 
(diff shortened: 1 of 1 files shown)


nuget/AppCenterCrashes.nuspec
```diff
@@ -47,6 +47,7 @@
     <!-- targets files add correct references to proper platforms -->
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.targets" target="build/uap10.0"/>
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.dll" target="ref/uap10.0" />
+    <file src="$uwp_dir$/x86/Microsoft.AppCenter.Crashes.xml" target="ref/uap10.0" />
```
guperrot: The x86 is weird but it works e2e when project configured on x64, I just tested 😅

Zongrenfeng: Yeah, since we don't know the correct location of the xml file exist on CI, I asked Dmitry's help for compare the files of uap and x86, and confirmed they are same. So this can be used as a work around.

PR #881, opened 2019-04-19
Add auth module with failing portable signIn

guperrot:

## Description

Add auth module with failing portable signIn

### DIFF
--- a/SDK/AppCenterAuth/Microsoft.AppCenter.Auth.Shared/SignInResult.cs
+++ b/SDK/AppCenterAuth/Microsoft.AppCenter.Auth.Shared/SignInResult.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.AppCenter.Auth
+{
+    /// <summary>
+    /// Result for sign-in operation. 
+    /// </summary>
+    public class SignInResult
+    {
+        /// <summary>
+        /// Gets the user information.
+        /// </summary>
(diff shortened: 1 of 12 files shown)


SDK/AppCenterAuth/Microsoft.AppCenter.Auth.Shared/SignInResult.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// Gets the user information.
+        /// </summary>
+        /// <value>The user information or null if an error occurred.</value>
+        public UserInformation UserInformation { get; internal set; }
+
+        /// <summary>
+        /// Error that occurred during sign-in.
+        /// </summary>
+        /// <value>The exception that describes the error or null if sign-in succeeded.</value>
+        public Exception Exception { get; internal set; }
```
MatkovIvan: It's already stored in `Task`

guperrot: awaiting would throw if using the task c# exception but our SDK does not throw in general.

guperrot: On second thought the argument for not throwing was because we were mostly an analytics/crash sdk at first (that can be removed without changing the behavior of the app).

For something that actually has a result and first class citizen dependency of the app that might make sense to throw, I'll speak again about that with the team and see what they think.

guperrot: Team agreed that on our C# SDK we should throw on signIn failure, so will make the change.

PR #883, opened 2019-04-19
Android Auth module

guperrot:

## Description

Auth.SignInAsync Android support.

[AB#60471](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/60471)


SDK/AppCenterAuth/Microsoft.AppCenter.Auth.Android/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,30 @@
...
+//
+//      Major Version
+//      Minor Version 
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers 
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
```
guperrot: need to fix

SDK/AppCenterAuth/Microsoft.AppCenter.Auth.Android/Resources/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<resources>
+    <string name="hello">Hello World, Click Me!</string>
+    <string name="app_name">Microsoft.AppCenter.Auth.Android</string>
```
guperrot: need to clean that

SDK/AppCenterAuth/Microsoft.AppCenter.Auth.Android/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,31 @@
...
+
+// General Information about an assembly is controlled through the following 
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTitle("Microsoft.AppCenter.Auth.Android")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("Microsoft.AppCenter.Auth.Android")]
+[assembly: AssemblyCopyright("Copyright ©  2018")]
```
MatkovIvan: Should be updated

SDK/AppCenterAuth/Microsoft.AppCenter.Auth.Android/Auth.cs
```diff
@@ -22,7 +35,8 @@ private static Task<UserInformation> PlatformSignInAsync()
                 var signInResult = (SignInResult)future.Get();
                 if (signInResult.Exception != null)
                 {
-                    throw signInResult.Exception;
+                    // Keep the stacktrace clean.
```
dhei: It would be nice to add a comment to note the behavior difference between native android and xamarin.android that one throws and one doesn't.

MatkovIvan: The behavior is the same - both throwing the exception. But iOS doesn't have C# `Exception` object to (re)throw, so we don't need to keep the previous stacktrace there.

dhei: @MatkovIvan Sorry I wan't clear here. I mean native android sign-in API which doesn't throw but holds the exception at `signInResult.Exception` that is different from xamarin android which throws exception on sign-in API.

guperrot: It's documented in the shared code that it throws.
We have shared code for all platforms that calls the same method name with Platform prefix.

dhei: Sounds good.

PR #892, opened 2019-04-23
Fix signIn on simulator

guperrot:

## Description

Fix signIn on simulator.


guperrot:
Without keychain capability, MSAL returns null on homeAccountId, accountId and idToken, which the native AppCenter SDK does not check for and think it's a success.

We also need to file a bug in native SDK to make an error for that case.

guperrot:
Native SDK is actually triggering an error for that case:
```
2019-04-22 19:19:51.475973-0700 SasquatchPuppet[68104:3884910] [AppCenterIdentity] ERROR: -[MSIdentity acquireTokenInteractively]_block_invoke/382 User sign-in failed. Error: Error Domain=MSIDKeychainErrorDomain Code=-34018 "(null)" UserInfo={MSIDCorrelationIdKey=7D285A09-E612-452F-A27D-ABE9181F20B4, MSIDErrorDescriptionKey=Failed to set item into keychain.}
```
So will test for null in Xamarin instead as Xamarin seems to change the behavior.

Issue #821, opened 2019-01-28
Add Tuple based overloads to Analytics/Crashes


dansiegel:
### **Description**

There are a number of times where a developer may only want to add a single property to track such as a UserId. Given the current API's you have to instantiate a Dictionary and load it like:

```cs
Analytics.TrackEvent("Some Event", new Dictionary<string, string> { { "userid", userId } });
```

### Proposed API

- Analytics.TrackEvent(string message, params (string, string)[] properties)
- Crashes.TrackError(Exception ex, params (string, string)[] properties)

These overloaded API's don't actually change or impact functionality in any way, but they do make it just a little easier and provide some nice syntactic sugar when calling them.

```cs
Analytics.TrackEvent("Some Event", ("userId", userId));
```

guperrot:
Hi, thanks for your feedback, I will forward it.
As a side note, for userId we recommend using AppCenter.SetUserId instead.

dansiegel:
@guperrot I was just using that as a common case that would make sense for someone to want to track... and yes the SetUserId would seem like a much better API to use for that specific property. :)

guperrot:
We might consider this feature in the future but there is no plan right now.

guperrot:
Sure, for Analytics you need to modify portable code (empty implementation): https://github.com/Microsoft/AppCenter-SDK-DotNet/blob/develop/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics/Analytics.cs

Then implement it for Android, iOS then Windows.

https://github.com/Microsoft/AppCenter-SDK-DotNet/blob/develop/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.iOS/Analytics.cs
https://github.com/Microsoft/AppCenter-SDK-DotNet/blob/develop/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Android/Analytics.cs
https://github.com/Microsoft/AppCenter-SDK-DotNet/blob/develop/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Analytics.cs

For crashes it can be done in a single place as code is shared: https://github.com/Microsoft/AppCenter-SDK-DotNet/blob/develop/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared/Crashes.cs

fredeil:
Would it be ok introduce the System.ValuTuple NuGet into the project? @guperrot

[...]

PR #951, opened 2019-05-18
Analytics error tracking key/value

fredeil:
## Description

Add support for string overloads on TrackEvent and TrackError. Decided to not go for a valueTuple since it introduces an unnecessary dependency.

## Related PRs or issues

https://github.com/microsoft/appcenter-sdk-dotnet/issues/821

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter/Microsoft.AppCenter.csproj
+++ b/SDK/AppCenter/Microsoft.AppCenter/Microsoft.AppCenter.csproj
@@ -15,7 +15,6 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="MSBuild.Sdk.Extras" Version="1.1.0" PrivateAssets="All" />
-    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
   </ItemGroup>
   <Import Project="$(MSBuildSDKExtrasTargets)" Condition="Exists('$(MSBuildSDKExtrasTargets)')" />
   <Target Name="GetPackagingOutputs" />
--- a/SDK/AppCenter/Microsoft.AppCenter/Microsoft.AppCenter.csproj
+++ b/SDK/AppCenter/Microsoft.AppCenter/Microsoft.AppCenter.csproj
@@ -15,6 +15,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="MSBuild.Sdk.Extras" Version="1.1.0" PrivateAssets="All" />
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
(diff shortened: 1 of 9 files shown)


fredeil:
Could someone give me details on the build?

cc @jwallra

SDK/AppCenter/Microsoft.AppCenter/Microsoft.AppCenter.csproj
```diff
@@ -15,6 +15,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="MSBuild.Sdk.Extras" Version="1.1.0" PrivateAssets="All" />
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
guperrot: This new dependency needs to be discussed.

jwallra:
@fredeil I wasn't sure either, it seemed like it should be included in .NET, but it appears that it is only included by default after .NET 4.7, .NET Standard 2.0 and .NET Core 2.0.  See this stack overflow question: https://stackoverflow.com/questions/38382971/predefined-type-system-valuetuple%C2%B42%C2%B4-is-not-defined-or-imported

Since it is an external dependency we need to determine whether we want to take it or not

fredeil:
Ok, maybe we could satisfy #821 by providing a string/string overloqd, like:

«void TrackEvent(string name, string key, string value)»

SDK/AppCenter/Microsoft.AppCenter/Microsoft.AppCenter.csproj
guperrot: Also that's the cause of the build failure. It's not referenced correctly everywhere it's needed.

guperrot:
> Ok, maybe we could satisfy #821 by providing a string/string overloqd, like:
> 
> «void TrackEvent(string name, string key, string value)»

Sure you can do that solution instead of the tuples.

fredeil:
The tuples are removed now, System.ValueTuple is also gone now. Waiting for build results @jwallra @guperrot

jwallra:
We're looking at this change now.  We're considering if we were to accept it would we need to implement something similar on other SDKs for consistency.

guperrot:
The consistency with all other SDKs is being discussed.
We will also have a consistency issue with `Crashes.TrackError` and in other SDKs other APIs that take a dictionary.

fredeil:
Ok, let me know what the outcome is

guperrot:
Sorry that I didn't came back here earlier.
There is no plan for now to modify the signature on all SDKs.
Since the PR is linked to the feature request issue (which remains open), it won't be lost, so I can close here for now.

PR #956, opened 2019-06-04
User info UI

Jamminroot:
This PR adds user info UI to Xamarin.Forms

## Description
UI demo

## Related PRs or issues

[AB#62368](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/62368)

### DIFF
--- a/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml
+++ b/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml
@@ -21,7 +21,7 @@
                 <Button Text="Sign Out" InputTransparent="true" Clicked="SignOut" />
             </ViewCell>
             <ViewCell Tapped="SignInInformation">
-                <Button Text="Signed out" InputTransparent="true" Clicked="SignInInformation" />
+                <Button x:Name="SignInInformationButton" Text="User Not Authenticated" InputTransparent="true" Clicked="SignInInformation" />
             </ViewCell>
         </TableSection>
     </TableView>
--- a/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml
+++ b/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml
@@ -20,6 +20,9 @@
             <ViewCell Tapped="SignOut">
                 <Button Text="Sign Out" InputTransparent="true" Clicked="SignOut" />
             </ViewCell>
+            <ViewCell Tapped="SignInInformation">
+                <Button Text="Signed out" InputTransparent="true" Clicked="SignInInformation" />
+            </ViewCell>
         </TableSection>
     </TableView>
 </ContentPage>
--- a/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml
+++ b/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml
@@ -21,7 +21,7 @@
                 <Button Text="Sign Out" InputTransparent="true" Clicked="SignOut" />
             </ViewCell>
             <ViewCell Tapped="SignInInformation">
-                <Button x:Name="SignInInformationButton" Text="User Not Authenticated" InputTransparent="true" Clicked="SignInInformation" />
+                <Button x:Name="SignInInformationButton" Text="Unknown" InputTransparent="true" Clicked="SignInInformation" />
             </ViewCell>
         </TableSection>
     </TableView>
--- a/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
+++ b/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
@@ -99,7 +99,8 @@ namespace Contoso.Forms.Puppet
         {
             try
             {
-                await Auth.SignInAsync().ContinueWith(task => {
+                await Auth.SignInAsync().ContinueWith(task =>
+                {
                     userInfo = task.Result;
                     SignInInformationButton.Text = "User authenticated";
                 });
@@ -148,16 +149,16 @@ namespace Contoso.Forms.Puppet
                     SomeNumber = 123,
                     SomeObject = dict,
                     SomePrimitiveArray = new int[] { 1, 2, 3 },
-                    SomeObjectArray = new CustomDocument[] { 
-                        new CustomDocument { 
+                    SomeObjectArray = new CustomDocument[] {
+                        new CustomDocument {
                             Id = Guid.NewGuid(),
                             TimeStamp = DateTime.UtcNow,
                             SomeNumber = 123,
                             SomeObject = dict,
                             SomePrimitiveArray = new int[] { 1, 2, 3 },
                             SomeObjectCollection = objectCollection,
-                            SomePrimitiveCollection = primitiveCollection  
-                        } 
+                            SomePrimitiveCollection = primitiveCollection
+                        }
                     },
                     SomeObjectCollection = objectCollection,
                     SomePrimitiveCollection = primitiveCollection,
@@ -218,7 +219,7 @@ namespace Contoso.Forms.Puppet
 
         async void SignInInformation(object sender, EventArgs e)
         {
-            if (userInfo!=null)
+            if (userInfo != null)
             {
                 string accessToken = Decode(userInfo.AccessToken);
                 string idToken = Decode(userInfo.IdToken);
--- a/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
+++ b/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
@@ -196,27 +196,12 @@ namespace Contoso.Forms.Puppet
             SignInInformationButton.Text = "User not authenticated";
         }
 
-        string ObfuscateToken(string token)
-        {
-            if (token?.Length == 0) 
-            {
-                return token;
-            }
-            var tokenData = token.Split('.');
-            if (tokenData.Length != 3) 
-            {
-                return token;
-            }
-            tokenData[2] = "***";
-            return String.Join(".", tokenData);
-        }
-
         async void SignInInformation(object sender, EventArgs e)
         {
             if (userInfo != null)
             {
-                string accessToken = ObfuscateToken(userInfo.AccessToken);
-                string idToken = ObfuscateToken(userInfo.IdToken);
+                string accessToken = userInfo.AccessToken?.Length > 0 ? "Set" : "Unset";
+                string idToken = userInfo.IdToken?.Length > 0 ? "Set" : "Unset";
                 await Navigation.PushModalAsync(new SignInInformationContentPage(userInfo.AccountId, accessToken, idToken));
             }
         }
--- a/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
+++ b/Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
@@ -59,7 +59,10 @@ namespace Contoso.Forms.Puppet
             RumEnabledSwitchCell.IsEnabled = acEnabled;
             EventFilterEnabledSwitchCell.On = _eventFilterStarted && await EventFilterHolder.Implementation?.IsEnabledAsync();
             EventFilterEnabledSwitchCell.IsEnabled = acEnabled && EventFilterHolder.Implementation != null;
-            if (userInfo?.AccountId != null) SignInInformationButton.Text = "User authenticated";
+            if (userInfo?.AccountId != null)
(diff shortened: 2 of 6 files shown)


Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -181,6 +188,46 @@ async void RunMBaaSAsync(object sender, EventArgs e)
         void SignOut(object sender, EventArgs e)
         {
             Auth.SignOut();
+            SignInInformationButton.Text = "User Not Authenticated";
+        }
+
+        private static byte[] Base64UrlDecode(string input)
```
MatkovIvan: https://stackoverflow.com/a/11743162/1703290

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -181,6 +188,46 @@ async void RunMBaaSAsync(object sender, EventArgs e)
...
+                case 0: break; // No pad chars in this case
+                case 2: output += "=="; break; // Two pad chars
+                case 3: output += "="; break; // One pad char
+                default: throw new System.Exception("Illegal base64url string!");
+            }
+            var converted = Convert.FromBase64String(output); // Standard base64 decoder
+            return converted;
+        }
+
+        public static string Decode(string token, string key)
```
MatkovIvan: Why doesn't `JwtSecurityTokenHandler` work here?

MatkovIvan: Maybe [Jwt.Net](https://www.nuget.org/packages/JWT) or [Portable.JWT](https://www.nuget.org/packages/Portable.JWT) ?

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml
```diff
@@ -20,6 +20,9 @@
             <ViewCell Tapped="SignOut">
                 <Button Text="Sign Out" InputTransparent="true" Clicked="SignOut" />
             </ViewCell>
+            <ViewCell Tapped="SignInInformation">
+                <Button x:Name="SignInInformationButton" Text="User Not Authenticated" InputTransparent="true" Clicked="SignInInformation" />
```
annakocheshkova: I prefer setting default value as "Unknown" until the first call to SignIn is made. Because user could actually be authenticated, we just don't know.

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -181,6 +188,46 @@ async void RunMBaaSAsync(object sender, EventArgs e)
...
+            byte[] headerData = Base64UrlDecode(header);
+            var headerJson = Encoding.UTF8.GetString(headerData,0, headerData.Length);
+            byte[] payloadData = Base64UrlDecode(payload);
+            var payloadJson = Encoding.UTF8.GetString(payloadData, 0, payloadData.Length);
+            return payloadJson;
+        }
+
+        async void SignInInformation(object sender, EventArgs e)
+        {
+            if (userInfo!=null)
```
annakocheshkova:
```suggestion
            if (userInfo != null)
```

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -181,6 +188,46 @@ async void RunMBaaSAsync(object sender, EventArgs e)
...
+            {
+                string accessToken = Decode(userInfo.AccessToken, "");
```
annakocheshkova: Why have the second parameter while not using it? You might as well just drop it since you are using a custom decoder anyway.

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
guperrot: No please use the Microsoft one we use in ingestion if possible, https://www.nuget.org/packages/System.IdentityModel.Tokens.Jwt/

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/SignInInformationContentPage.xaml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ContentPage xmlns="http://xamarin.com/schemas/2014/forms"
+             xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml" 
+             x:Class="Contoso.Forms.Puppet.SignInInformationContentPage">
+    <StackLayout Padding="20">
+        <StackLayout Orientation="Horizontal">
+            <Label Text="AccountId" HorizontalOptions="Start" VerticalOptions="Center" WidthRequest="120" />
+            <Entry HorizontalOptions="FillAndExpand" x:Name="AccountIdEntry" IsEnabled="false"/>
```
annakocheshkova: Why not Label?

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
annakocheshkova: By the way I may be wrong, but I got the impression that token can be null and it might express "unauthenticated" user.

guperrot:
As per @dougnutz comment on a direct message discussion. Decoding the token in the test app is out of scope of the feature requirements.

I also personally think we should not expose the raw token directly in logs or UI as its credentials.

To the very least we can obfuscate the value (like just truncating the end after the second dot which correspond to the signature part which is sensitive part) and just display what remains.

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
Jamminroot: MS one is incompatible. Requires .netstandart >= 1.4;
As for other dependencies - since the code for decoding is quite short, I decided not to add extra dependencies, and just code it in place.

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -55,6 +59,8 @@ protected override async void OnAppearing()
             RumEnabledSwitchCell.IsEnabled = acEnabled;
             EventFilterEnabledSwitchCell.On = _eventFilterStarted && await EventFilterHolder.Implementation?.IsEnabledAsync();
             EventFilterEnabledSwitchCell.IsEnabled = acEnabled && EventFilterHolder.Implementation != null;
+            if (userInfo?.AccountId != null) SignInInformationButton.Text = "User authenticated";
```
annakocheshkova: `{ }` please don't forget us! ☺️

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -95,7 +101,8 @@ async void RunMBaaSAsync(object sender, EventArgs e)
         {
             try
             {
-                var userInfo = await Auth.SignInAsync();
+                userInfo = await Auth.SignInAsync();
+                if (userInfo.AccountId != null) SignInInformationButton.Text = "User authenticated";
```
annakocheshkova: ☺️

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -181,6 +188,27 @@ async void RunMBaaSAsync(object sender, EventArgs e)
         void SignOut(object sender, EventArgs e)
         {
             Auth.SignOut();
+            userInfo = null;
+            SignInInformationButton.Text = "User not authenticated";
```
annakocheshkova: A constant?

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -181,6 +188,27 @@ async void RunMBaaSAsync(object sender, EventArgs e)
...
+        }
+
+        string ObfuscateToken(string token)
+        {
+            if (token?.Length == 0) return token;
```
jwallra:
```suggestion
            if (token?.Length == 0) 
            {
                return token;
            }
```

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
```diff
@@ -181,6 +188,27 @@ async void RunMBaaSAsync(object sender, EventArgs e)
...
+            var tokenData = token.Split('.');
+            if (tokenData.Length != 3) return token;
```
jwallra:
```suggestion
            if (tokenData.Length != 3) 
            {
                return token;
            }
```

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
jwallra:
```suggestion
                if (userInfo.AccountId != null)
                {
                    SignInInformationButton.Text = "User authenticated";
                }
```

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
jwallra:
```suggestion
            if (userInfo?.AccountId != null)
            {
                SignInInformationButton.Text = "User authenticated";
            }
```

Apps/Contoso.Forms.Puppet/Contoso.Forms.Puppet/ModulePages/OthersContentPage.xaml.cs
Jamminroot: Might have been made one, yeah. Didn't make one since other code don't tend to use constants for strings.
Guess it should be fixed

guperrot:
Also on iOS it depends on a weird state of the sources where we haven't modified MSUserInformation bindings yet so it does not even build (but on develop we have the properties).

guperrot:
After merging locally develop, iOS app works but is unreadable: notch in the way and no scrolling.
At this point since the last decision was to not display tokens at all but rather booleans if property set to not null or not, might be a simpler fix.

guperrot:
> Clicking on user authenticated on Android crashes the app:
> 
> ```
> 2019-06-05 14:22:51.922 5776-5776/com.microsoft.appcenter.xamarin.forms.puppet E/AppCenterCrashes: Unhandled Exception from source=AndroidEnvironment
>     System.NullReferenceException: Object reference not set to an instance of an object
>       at Contoso.Forms.Puppet.OthersContentPage.ObfuscateToken (System.String token) [0x00019] in <eeaaf087d9ea48a1b1caeee57cf61c04>:0 
>       at Contoso.Forms.Puppet.OthersContentPage.SignInInformation (System.Object sender, System.EventArgs e) [0x0003f] in <eeaaf087d9ea48a1b1caeee57cf61c04>:0 
>       at System.Runtime.CompilerServices.AsyncMethodBuilderCore+<>c.<ThrowAsync>b__6_0 (System.Object state) [0x00000] in <58604b4522f748968296166e317b04b4>:0 
>       at Android.App.SyncContext+<>c__DisplayClass2_0.<Post>b__0 () [0x00000] in <788a34f7a7b84486905dfde786529d42>:0 
>       at Java.Lang.Thread+RunnableImplementor.Run () [0x00008] in <788a34f7a7b84486905dfde786529d42>:0 
>       at Java.Lang.IRunnableInvoker.n_Run (System.IntPtr jnienv, System.IntPtr native__this) [0x00009] in <788a34f7a7b84486905dfde786529d42>:0 
>         at (wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.44(intptr,intptr)
> ```
> 
> As advised in chat channel today, if it's faster to switch to boolean (true if not null token, or false if null) then it might be a way to fix. We really need to close this soon.

Looks like  a build cache issue where the new properties were always null somehow, fixed after clean.

guperrot:
The team decided in standup that for unfinished work (so not Android which was closed), we won't display either token or parsed token but instead just whether the token is set in response or not.

Since the current PR contained some UI glitches (mostly on iOS) and had to be patched, applied the rule decided in standup to unfinished work item.

AccountId was also truncated, fixed that and the notch and the scrolling is non issue now since we don't display much now.

PR #966, opened 2019-06-12
Enable packaging wpf/winforms dlls for core/crashes / avoid crash for unsupported modules at start

guperrot:

## Description

Enable packaging wpf/winforms dlls for core/crashes.

## Related PRs or issues

[AB#62888](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/62888)

### DIFF
--- a/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics/Analytics.cs
+++ b/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics/Analytics.cs
@@ -9,7 +9,7 @@ namespace Microsoft.AppCenter.Analytics
     /// <summary>
     ///     Analytics feature.
     /// </summary>
-    public partial class Analytics
+    public class Analytics
     {
         private Analytics()
         {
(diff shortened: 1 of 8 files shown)


SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics/Analytics.cs
```diff
@@ -6,11 +6,10 @@
 
 namespace Microsoft.AppCenter.Analytics
 {
-    /// <inheritdoc />
     /// <summary>
     ///     Analytics feature.
     /// </summary>
-    public class Analytics : IAppCenterService
```
guperrot: Fixes crash when adding Analytics nuget to WPF or Winforms (since we don't include dll, portable dll was loaded, implementing an interface but no implementation for ServiceName => TypeLoadException).

guperrot: Works perfectly fine without that by the way, tested direct, and via .NET standard bait/switch (forms.puppet.uwp).

guperrot:
Tested with nugets in winforms demo.

PR #975, opened 2019-06-18
Feature/send crashes

achocron:

## Description

This PR addresses catching and saving crashes to disk. Sending them will be in forthcoming work.


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Ingestion/Models/Log.cs
```diff
@@ -1,18 +1,19 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
dhei: Please note that files with "Ingestoin/Models" in the path are auto generated by `autorest`.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Device.cs
```diff
@@ -0,0 +1,32 @@
...
+        public Device(Ingestion.Models.Device device)
+        {
+            SdkName = device.SdkName;
+            SdkVersion = device.SdkVersion;
+            Model = device.Model;
+            OemName = device.OemName;
+            OsName = device.OsName;
+            OsVersion = device.OsVersion;
+            OsBuild = device.OsBuild;
+            OsApiLevel = device.OsApiLevel ?? 0; //TODO verify
```
guperrot: Address TODO before merging

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/AbstractDeviceInformationHelper.cs
```diff
@@ -4,16 +4,20 @@
...
 using System.Threading.Tasks;
-
 using Microsoft.AppCenter.Windows.Shared.Utils;
 
 namespace Microsoft.AppCenter.Utils
 {
     public abstract class AbstractDeviceInformationHelper : IDeviceInformationHelper
     {
-        public virtual async Task<Ingestion.Models.Device> GetDeviceInformationAsync()
+        public virtual Task<Ingestion.Models.Device> GetDeviceInformationAsync()
```
guperrot: Why making Async synchronous in reality? There is no value in having the async since it will block before even returning a task object.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/FileHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// Utilities for the file system.
+    /// </summary>
+    public class FileHelper
+    {
+        private readonly DirectoryInfo _baseDirectory;
+
+        /// <summary>
+        /// Creates an instance. This parameterless constructor is needed for tests.
+        /// </summary>
+        public FileHelper()
```
guperrot: Have we considered using internal, and internalsVisibleTo to the test assembly

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/FileHelper.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <summary>
+        /// Creates an instance with the given base path.
+        /// </summary>
+        /// <param name="basePath">The path to use as the base for all operations.</param>
+        public FileHelper(string basePath)
+        {
+#if WINDOWS_UWP
+            var basePathPrefix = global::Windows.Storage.ApplicationData.Current.LocalFolder.Path;
+#else
+            //TODO what should this be?
```
guperrot: TODO

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.WindowsDesktop/Utils/ProcessInformation.cs
```diff
@@ -0,0 +1,81 @@
...
+        // Parent process information is not (easily) available, and not necessary.
+        public string ParentProcessName => null;
+
+        public string ProcessArchitecture
+        {
+            get
+            {
+                try
+                {
+                    return Environment.GetEnvironmentVariable("PROCESSOR_ARCHITECTURE"); // TODO improve
```
guperrot: TODO

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.WindowsDesktop/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Newtonsoft.Json" version="6.0.1" targetFramework="net45" />
+</packages>
```
guperrot: I cannot comment on binaries itself so doing it in the file just before.

Why do we commit binaries to git?

* Tests/Microsoft.AppCenter.Crashes.Test.Windows/Fakes/System.fakes
* Tests/Microsoft.AppCenter.Crashes.Test.Windows/Fakes/mscorlib.fakes

dhei: @guperrot 

GitHub doesn't recognize non standard `UTF-16` encoding.

`*.fakes` are text files encoded in `UTF-16 LE`, a special `UTF-16` with BOM according to Wikipedia (https://en.wikipedia.org/wiki/UTF-16).

> The standard also allows the byte order to be stated explicitly by specifying UTF-16BE or UTF-16LE as the encoding type. When the byte order is specified explicitly this way, a BOM is specifically not supposed to be prepended to the text, and a U+FEFF at the beginning should be handled as a ZWNBSP character. Most applications ignore a BOM in all cases despite this rule.

image

guperrot: Oh wow.
Ok 😅

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/Exception.cs
```diff
@@ -0,0 +1,102 @@
...
+        public string StackTrace { get; set; }
+
+        /// <summary>
+        /// Gets or sets inner exceptions of this exception.
+        /// </summary>
+        [JsonProperty(PropertyName = "innerExceptions")]
+        public IList<Exception> InnerExceptions { get; set; }
+
+        /// <summary>
+        /// Gets or sets name of the wrapper SDK that emitted this exeption.
```
michaelsmitha:
```suggestion
        /// Gets or sets name of the wrapper SDK that emitted this exception.
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/Exception.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// Consists of the name of the SDK and the wrapper platform, e.g.
+        /// "appcenter.xamarin", "hockeysdk.cordova".
+        ///
```
michaelsmitha:
```suggestion
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/Exception.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "wrapperSdkName")]
+        public string WrapperSdkName { get; set; }
+
+        /// <summary>
+        /// Validate the object.
+        /// </summary>
+        /// <exception cref="ValidationException">
+        /// Thrown if validation fails
```
michaelsmitha:
```suggestion
        /// Thrown if validation fails.
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/ManagedErrorLog.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// </param>
+        /// <param name="sid">When tracking an analytics session, logs can be
+        /// part of the session by specifying this identifier.
+        /// This attribute is optional, a missing value means the session
+        /// tracking is disabled (like when using only error reporting
+        /// feature).
+        /// Concrete types like StartSessionLog or PageLog are always part of a
+        /// session and always include this identifier.
+        /// </param>
+        /// <param name="userId">optional string used for associating logs with
```
michaelsmitha:
```suggestion
        /// <param name="userId">Optional string used for associating logs with
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/ManagedErrorLog.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        /// <summary>
+        /// Gets or sets unique ID for a Xamarin build or another similar
+        /// technology.
+        /// </summary>
+        [JsonProperty(PropertyName = "buildId")]
+        public string BuildId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the exception
```
michaelsmitha:
```suggestion
        /// Gets or sets the exception.
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/ManagedErrorLog.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "exception")]
+        public Exception Exception { get; set; }
+
+        /// <summary>
+        /// Validate the object.
+        /// </summary>
+        /// <exception cref="ValidationException">
+        /// Thrown if validation fails
```
michaelsmitha:
```suggestion
        /// Thrown if validation fails.
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.WindowsDesktop/Utils/ProcessInformation.cs
achocron: Yep, known TODO - which is why we have "do not merge"

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/FileHelper.cs
achocron: Yep, known TODO - which is why we have "do not merge"

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/AbstractDeviceInformationHelper.cs
jwallra: Also if we do go through with this change, the method name needs to change.

achocron:
@michaelsmitha I was about to merge all of your suggestions, but I then realized that they are on files that have been automatically generated. Thus, the changes will always show up when we regenerate models. So I am wondering if we should leave them.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/AbstractDeviceInformationHelper.cs
achocron: No this returns a `Task` so nothing actually changes. It won't block before returning a `Task`. Have you looked at the UWP implementation? It is the reason we need to return a `Task`. But since this base implementation does not `await`, using `async` doesn't do anything.

achocron: And it is still conventional to use the `Async` suffix when you return a `Task`, even if the keyword is not used. We also introduced a truly synchronous version that is needed for the crash module. That one doesn't block at all.

michaelsmitha:
@achocron That's fine. I wasn't sure where the generated things were pulling from. Maybe if the right ones are merged then the lines that are off/different than the others will be generated with the fixes? Do what you think is best.

achocron:
The generated code is based on the ingestion service's swagger file, and AutoRest itself. But most of the comments come from the swagger file. I think the documentation there can be improved, but that's in another repository. I'm going to resolve the feedbacks without changing the comments, but if you insist that we fix the documentation, I don't mind changing it.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/AbstractErrorLog.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// </param>
+        /// <param name="sid">When tracking an analytics session, logs can be
+        /// part of the session by specifying this identifier.
+        /// This attribute is optional, a missing value means the session
+        /// tracking is disabled (like when using only error reporting
+        /// feature).
+        /// Concrete types like StartSessionLog or PageLog are always part of a
+        /// session and always include this identifier.
+        /// </param>
+        /// <param name="userId">optional string used for associating logs with
```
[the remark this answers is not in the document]
jwallra: Files in Ingestion/Models are autogenerated ;)  I assume that these will need to be changed, but somewhere else.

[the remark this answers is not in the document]
dhei: https://github.com/microsoft/appcenter-sdk-dotnet/pull/975#discussion_r295067285

Let's not change the autogenerated code.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.WindowsDesktop/Utils/ProcessInformation.cs
```diff
@@ -0,0 +1,81 @@
...
+                }
+                catch (Exception e)
+                {
+                    AppCenterLog.Warn(Crashes.LogTag, "Unable to get process name.", e);
+                    return null;
+                }
+            }
+        }
+
+        // Parent process information is not (easily) available, and not necessary.
```
jwallra: Can you give a bit more context on this?  Did the crashes team say it's not necessary?  How would we get the parent process information if we decided we want it?

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
```diff
@@ -0,0 +1,240 @@
...
+                try
+                {
+                    // Convert to list so enumeration does not occur outside the lock.
+                    return FileHelper.EnumerateFiles($"*{ErrorLogFileExtension}").ToList();
+                }
+                catch (System.Exception ex)
+                {
+                    AppCenterLog.Error(Crashes.LogTag, "Failed to retrieve error log files.", ex);
+                }
+                return null;
```
jwallra: Should return empty list

```suggestion
                return new List<FileInfo>();
```

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/AbstractDeviceInformationHelper.cs
guperrot: We need to discuss it offline.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/FileHelper.cs
guperrot: This is to make sure its tracked as a requested change before merge, todo are often left out ;)

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.WindowsDesktop/Utils/ProcessInformation.cs
guperrot: This is to make sure its tracked as a requested change before merge, todo are often left out ;)

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -43,6 +52,7 @@ public static Crashes Instance
 
         public void OnChannelGroupReady(IChannelGroup channelGroup, string appSecret)
         {
+            PlatformSetEnabledAsync(true);
```
jaelim-ms: Why do you call this one here?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/AbstractDeviceInformationHelper.cs
guperrot: So in UWP we override it to be real async, ok.

Since we have only 2 implementations it should really be only in windows desktop the task.fromresult, that would make the whole thing less confusing. So please make method abstract and implement only task.fromresult in windowsdesktop.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/IDeviceInformationHelper.cs
```diff
@@ -11,9 +11,15 @@ namespace Microsoft.AppCenter.Utils
...
+        /// Gets the device information asynchronously.
         /// </summary>
         /// <returns>Device object with fields populated appropriately</returns>
         Task<Ingestion.Models.Device> GetDeviceInformationAsync();
+
+        /// <summary>
+        /// Gets the device information without blocking. This might be missing certain properties.
+        /// </summary>
+        /// <returns>Device object with fields populated appropriately</returns>
+        Ingestion.Models.Device GetDeviceInformation();
```
guperrot: This method does not need to be in the interface. The code in this branch always access it without the interface and thus resharper complains.

Either use interface or remove the method from the interface to resolve the warning.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/Constants.cs
```diff
@@ -17,5 +17,8 @@ public static class Constants
         public const int DefaultTriggerCount = 50;
         public static readonly TimeSpan DefaultTriggerInterval = TimeSpan.FromSeconds(3);
         public const int DefaultTriggerMaxParallelRequests = 3;
+
+        // File Storage
```
guperrot: We use dots at end of comments, please fix that in at least the new or modified comments in this PR.

guperrot:
```suggestion
        // File Storage.
```

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/Constants.cs
```diff
@@ -0,0 +1,16 @@
...
+using System;
+
+namespace Microsoft.AppCenter.Utils
+{
+    /// <summary>
+    /// Various constants used by the SDK.
+    /// </summary>
+    public static partial class Constants
+    {
+        // File Storage
```
guperrot:
```suggestion
        // File Storage.
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
```diff
@@ -0,0 +1,246 @@
...
+            var fileName = errorLog.Id + ErrorLogFileExtension;
+            try
+            {
+                lock (LockObject)
+                {
+                    if (!CrashesDirectory.Exists)
+                    {
+                        CrashesDirectory.Create();
+                    }
+                    var filePath = Path.Combine(CrashesDirectory.FullName, fileName);
```
jaelim-ms: L160 and L161 doesn't need to be in the lock block.

jaelim-ms: Is your concern to store same error group ID?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/AbstractDeviceInformationHelper.cs
achocron: Sounds good. But let's do that in another PR

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.WindowsDesktop/Utils/ProcessInformation.cs
achocron: Yeah they said it is not necessary. The steps I believe involve using C++ code which we really want to avoid.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
achocron: It is just a naive way to get crashes started, the PR to do the enable/disable logic will overwrite it.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
achocron: Looks like addressed in the other PR.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/IDeviceInformationHelper.cs
achocron: The method is accessed on the interface in line 69 of ErrorLogHelper.cs. Maybe Resharper is confused since it is across multiple projects?

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
achocron: The concern is that the file can be deleted somehow - but I guess that is not a big deal because we would just catch the exception. I can move it.

SDK/AppCenter/Microsoft.AppCenter.UWP/Utils/DeviceInformationHelper.cs
```diff
@@ -16,20 +16,35 @@ public class DeviceInformationHelper : AbstractDeviceInformationHelper
     {
         public static event EventHandler InformationInvalidated;
         private static string _country;
-        private readonly IScreenSizeProvider _screenSizeProvider;
-        private static IScreenSizeProviderFactory _screenSizeProviderFactory =
+        private static IScreenSizeProvider ScreenSizeProvider;
+        private static IScreenSizeProviderFactory ScreenSizeProviderFactory =
```
dhei: Naming convention for `private static` property should be `_screenSizeProvider ` and `_screenSizeProviderFactory`.

PR #1011, opened 2019-06-27
Fix/error folder create

michaelsmitha:

## Description

Android uses a function to get the error storage directory and creates it on demand if needed. This should do the same.

## Related PRs or issues

[AB#63801](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/63801)

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
@@ -101,7 +101,7 @@ namespace Microsoft.AppCenter.Crashes.Utils
         public static File GetLastErrorLogFile() => Instance.InstanceGetLastErrorLogFile();
 
         /// <summary>
-        /// Creates the error storage directory if it does not exist.
+        /// Get the error storage directory, or creates it if it does not exist.
         /// </summary>
         /// <returns>The error storage directory.</returns>
         public static Directory GetErrorStorageDirectory() => Instance.InstanceGetErrorStorageDirectory();
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
@@ -226,21 +226,13 @@ namespace Microsoft.AppCenter.Crashes.Utils
         }
 
         /// <summary>
-        /// Creates the error storage directory if it does not exist.
+        /// Get the error storage directory, or creates it if it does not exist.
         /// </summary>
         /// <returns>The error storage directory.</returns>
         public virtual Directory InstanceGetErrorStorageDirectory()
         {
-            if (!_crashesDirectory.Exists())
-            {
-                lock (LockObject)
-                {
-                    if (!_crashesDirectory.Exists())
-                    {
-                        _crashesDirectory.Create();
-                    }
-                }
-            }
+
+            _crashesDirectory.Create();
             return _crashesDirectory;
         }
 
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
@@ -101,6 +101,12 @@ namespace Microsoft.AppCenter.Crashes.Utils
         public static File GetLastErrorLogFile() => Instance.InstanceGetLastErrorLogFile();
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
```diff
@@ -219,6 +225,25 @@ public virtual ManagedErrorLog InstanceReadErrorLogFile(File file)
...
+        /// Creates the error storage directory if it does not exist.
+        /// </summary>
+        /// <returns>The error storage directory.</returns>
+        public virtual Directory InstanceGetErrorStorageDirectory()
+        {
+            if (!_crashesDirectory.Exists())
+            {
+                lock (LockObject)
+                {
+                    if (!_crashesDirectory.Exists())
```
guperrot: We don't need to check twice, actually Create is already checking, also no need to lock.
I think this was already discussed with @achocron  on another task when we looked at the code of Create.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
```diff
@@ -100,6 +100,12 @@ public ErrorLogHelper()
         /// <returns>The most recently modified error log file.</returns>
         public static File GetLastErrorLogFile() => Instance.InstanceGetLastErrorLogFile();
 
+        /// <summary>
+        /// Creates the error storage directory if it does not exist.
```
guperrot: The getter talks only about create, which is unusual for a getter description. I think we need to rework the wording here.

jwallra:
```suggestion
        /// Get the error storage directory, or creates it if it does not exist.
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
jwallra: Yeah, this is a strange pattern.  I'd be surprised if we needed to check twice.  I would think that the outermost check could be removed.

achocron: The double-checking pattern saves us from locking unnecessarily so I see the motivation there. But as @guperrot said, we probably don't actually need to check even once. It's fine to just call Create without checking.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/Utils/ErrorLogHelperTest.cs
```diff
@@ -140,6 +140,37 @@ public void GetSingleErrorLogFileDoesNotThrow(Type exceptionType)
...
+            Mock.Get(mockDirectory).Verify(d => d.Create());
+        }
+
+        [TestMethod]
+        public void getErrorStorageDirectoryExists()
+        {
+            // Mock where directory doesn't exist.
+            var mockDirectory = Mock.Of<Directory>();
+            ErrorLogHelper.Instance._crashesDirectory = mockDirectory;
+            Mock.Get(mockDirectory).Setup(d => d.Create()).Throws(new System.Exception("Should not be called"));
```
jwallra: Won't this crash the test?  Where does the exception get caught?  I don't see it in the modified code

michaelsmitha: @jwallra The idea is it won't be called if it already exists. Since `Create` is already checking, I'll be removing this test anyway I think.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
michaelsmitha: Thanks for the input guys :) I learned that pattern from @achocron , as he said :)

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
```diff
@@ -100,6 +100,12 @@ public ErrorLogHelper()
         /// <returns>The most recently modified error log file.</returns>
         public static File GetLastErrorLogFile() => Instance.InstanceGetLastErrorLogFile();
 
+        /// <summary>
+        /// Get the error storage directory, or creates it if it does not exist.
```
jwallra:
```suggestion
        /// Gets the error storage directory, or creates it if it does not exist.
```

PR #1013, opened 2019-06-27
Add error attachment log model and tests

achocron:

## Description

Adds error attachment models.

[AB#63630](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/63630)

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
@@ -1,9 +1,9 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Text;
 using Microsoft.AppCenter.Ingestion.Models;
 using Newtonsoft.Json;
-using System.Text;
 
 namespace Microsoft.AppCenter.Crashes
 {
@@ -17,12 +17,20 @@ namespace Microsoft.AppCenter.Crashes
 
         static ErrorAttachmentLog PlatformAttachmentWithText(string text, string fileName)
         {
+            if (text == null)
+            {
+                return null;
+            }
             var data = Encoding.UTF8.GetBytes(text);
             return PlatformAttachmentWithBinary(data, fileName, ContentTypePlainText);
         }
 
         static ErrorAttachmentLog PlatformAttachmentWithBinary(byte[] data, string fileName, string contentType)
         {
+            if (data == null)
+            {
+                return null;
+            }
             return new ErrorAttachmentLog()
             {
                 Data = data,
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
@@ -1,18 +1,83 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using Microsoft.AppCenter.Ingestion.Models;
+using Newtonsoft.Json;
+using System.Text;
+
 namespace Microsoft.AppCenter.Crashes
 {
-    public partial class ErrorAttachmentLog
+    /// <summary>
+    /// Error attachment log.
+    /// </summary>
+    [JsonObject("errorAttachment")]
+    public partial class ErrorAttachmentLog : Log
     {
+        private const string ContentTypePlainText = "text/plain";
+
         static ErrorAttachmentLog PlatformAttachmentWithText(string text, string fileName)
         {
-            return new ErrorAttachmentLog();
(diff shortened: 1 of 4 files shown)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
```diff
@@ -1,18 +1,83 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using Microsoft.AppCenter.Ingestion.Models;
+using Newtonsoft.Json;
+using System.Text;
```
jaelim-ms: Sort?

Tests/Microsoft.AppCenter.Crashes.Test.Windows/ErrorAttachmentLogTest.cs
```diff
@@ -0,0 +1,41 @@
...
+            Assert.AreEqual(fileName, attachment.FileName);
+            CollectionAssert.AreEqual(data, attachment.Data);
+        }
+
+        [TestMethod]
+        public void TestCreateTextAttachmentLog()
+        {
+            // Create an error attachment log with text.
+            var text = "This is the text.";
+            var fileName = "binary attachment";
```
jaelim-ms: Is file name "binary attachment"? :trollface:

Tests/Microsoft.AppCenter.Crashes.Test.Windows/ErrorAttachmentLogTest.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Microsoft.AppCenter.Crashes.Test
+{
+    [TestClass]
+    public class ErrorAttachmentLogTest
```
jaelim-ms: Please add a test that doesn't set file name.

jaelim-ms: Please also add a test for Validation.

guperrot:
(please ignore the PR link, that was a mistake)

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
```diff
@@ -1,18 +1,83 @@
...
+
+        /// <summary>
+        /// Validate the object.
+        /// </summary>
+        /// <exception cref="ValidationException">
+        /// Thrown if validation fails
+        /// </exception>
+        public override void Validate()
+        {
+            base.Validate();
```
jwallra: this line should probably be last because the other ones will fail faster

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
```diff
@@ -1,18 +1,83 @@
...
+    /// </summary>
+    [JsonObject("errorAttachment")]
+    public partial class ErrorAttachmentLog : Log
     {
+        private const string ContentTypePlainText = "text/plain";
+
         static ErrorAttachmentLog PlatformAttachmentWithText(string text, string fileName)
         {
-            return new ErrorAttachmentLog();
+            var data = Encoding.UTF8.GetBytes(text);
```
clpolet: Shouldn't we check for null text? [Encoding.UTF8.GetBytes](https://docs.microsoft.com/en-us/dotnet/api/system.text.encoding.getbytes?view=netframework-4.8#System_Text_Encoding_GetBytes_System_String_) will throw an exception if `text` is null.

achocron: Yeah it I added the null check on another branch. Should appear here shortly.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
achocron: Copied straight from the autogenerated code so leaving as is.

jaelim-ms: Usually `base` or `super` is called at the beginning of a method when it is overridden. This is right order unless we have any other reasons but it wouldn't fail faster even though we move this down to the bottom in case base class properties are not valid.

achocron:
> Should have unit tests for error attachments in crashes. Is it coming in another PR?

@jaelim-ms There is no change to crashes in this PR, so I didn't add tests.

achocron:
@jaelim-ms Those changes were not supposed to be in this PR. Removed.

PR #1016, opened 2019-06-28
Feature/windows send error attachments

achocron:

## Description

Send error attachments if available for a crash.

Note: The part where the `ErrorReport` is created will likely be changed in the future (when we have exception available and probably some caching mechanism), but to avoid conflicts and double work, I kept it as simple as possible for now. (Also there is no exception file saved at the moment so calling the constructor is the only option anyways.)

[AB#63630](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/63630)

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
@@ -213,11 +213,14 @@ namespace Microsoft.AppCenter.Crashes
                 tasks.Add(Channel.EnqueueAsync(log));
                 _unprocessedManagedErrorLogs.Remove(key);
                 ErrorLogHelper.RemoveStoredErrorLogFile(key);
-                var errorReport = new ErrorReport(log, null);
+                if (GetErrorAttachments != null)
+                {
+                    var errorReport = new ErrorReport(log, null);
 
-                // This must never called while a lock is held.
-                var attachments = GetErrorAttachments?.Invoke(errorReport);
-                tasks.Add(SendErrorAttachmentsAsync(log.Id, attachments));
+                    // This must never called while a lock is held.
+                    var attachments = GetErrorAttachments?.Invoke(errorReport);
+                    tasks.Add(SendErrorAttachmentsAsync(log.Id, attachments));
+                }
             }
             return Task.WhenAll(tasks);
         }
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
@@ -214,23 +214,17 @@ namespace Microsoft.AppCenter.Crashes
                 tasks.Add(Channel.EnqueueAsync(log));
                 _unprocessedManagedErrorLogs.Remove(key);
                 ErrorLogHelper.RemoveStoredErrorLogFile(key);
-                
-                // Save callback into a local variable to avoid a race condition.
-                var errorAttachmentsCallback = GetErrorAttachments;
-                if (errorAttachmentsCallback != null)
-                {
-                    var errorReport = new ErrorReport(log, null);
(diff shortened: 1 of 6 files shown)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -213,9 +214,14 @@ private Task HandleUserConfirmationAsync()
...
                 ErrorLogHelper.RemoveStoredErrorLogFile(key);
-                var errorReport = new ErrorReport(log, null);
-                var attachments = GetErrorAttachments?.Invoke(errorReport);
-                tasks.Add(SendErrorAttachmentsAsync(log.Id, attachments));
+                if (GetErrorAttachments != null)
+                {
+                    var errorReport = new ErrorReport(log, null);
+
+                    // This must never called while a lock is held.
+                    var attachments = GetErrorAttachments?.Invoke(errorReport);
```
jaelim-ms: Redundant.
```suggestion
                    var attachments = GetErrorAttachments.Invoke(errorReport);
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
```diff
@@ -10,9 +10,11 @@ namespace Microsoft.AppCenter.Crashes
     /// <summary>
     /// Error attachment log.
     /// </summary>
-    [JsonObject("errorAttachment")]
+    [JsonObject(JsonIdentifier)]
```
jaelim-ms: Do we really need a variable for this one?

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
achocron: Not redundant because it can be unset by the time line 222 is reached

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
achocron: Yeah it's accessed in the Crashes class. This is the pattern used for all logs.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
jaelim-ms: Then why do you check it in L217?  You can get errorReport and get attachment, only add a task when attachment isn't null.

achocron: Because there is no need to create an error report unless the callback is set. The race condition is not as harmful in the other direction, just trying to avoid a crash. I think I have a better way that will be more clear though.

achocron: Improved.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -190,25 +194,77 @@ private Task ProcessPendingErrorsAsync()
...
+                
+                // Save callback into a local variable to avoid a race condition.
+                var errorAttachmentsCallback = GetErrorAttachments;
+                if (errorAttachmentsCallback != null)
+                {
+                    var errorReport = new ErrorReport(log, null);
+
+                    // This must never called while a lock is held.
+                    var attachments = errorAttachmentsCallback.Invoke(errorReport);
+                    tasks.Add(SendErrorAttachmentsAsync(log.Id, attachments));
```
jaelim-ms: Checking empty attachments here before creating tasks would help to reduce producing the number of Tasks.

PR #1020, opened 2019-07-03
Don't create empty exception file on UWP

guperrot:

## Description

Don't create empty exception file on UWP.

## Related PRs or issues

[AB#63628](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/63628)

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
@@ -216,7 +216,7 @@ namespace Microsoft.AppCenter.Crashes
                     AppCenterLog.Debug(LogTag, $"Process pending error file {file.Name}");
                     var log = ErrorLogHelper.ReadErrorLogFile(file);
 
-                    // Finish processing the file.
+                    // Delete file if corrupted.
                     if (log == null)
                     {
                         AppCenterLog.Error(LogTag, $"Error parsing error log. Deleting invalid file: {file.Name}");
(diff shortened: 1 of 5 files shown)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -236,27 +236,29 @@ private Task ProcessPendingErrorsAsync()
                         }
                         continue;
                     }
-                    var report = BuildErrorReport(log);
-                    if (report == null)
+
+                    // The error report cannot be built if any of those fields are null.
+                    if (log.Device == null || log.AppLaunchTimestamp == null || log.Timestamp == null)
```
clpolet: Would calling Validate() on the log make sense here?

jaelim-ms: It is a generated data model I guess.

jaelim-ms: [Here](https://github.com/microsoft/appcenter-sdk-dotnet/blob/develop/scripts/generated/Models/Log.cs)
In the generated model, `Device` is the only one nonnull.

jaelim-ms: I also had a same doubt but I believed my assumption is true. 😄

PR #1021, opened 2019-07-05
Error attachments aren't sent

AnastasiaKubova:

## Description

Text attachment error and file attachment error aren't sending after the crash. 

## Related PRs or issues

[AB#64184](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/64184)

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
@@ -356,7 +356,6 @@ namespace Microsoft.AppCenter.Crashes
                 {
                     attachment.Id = Guid.NewGuid();
                     attachment.ErrorId = errorId;
-
                     if (!attachment.ValidatePropertiesForAttachment())
                     {
                         AppCenterLog.Error(LogTag, "Not all required fields are present in ErrorAttachmentLog.");
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
@@ -358,7 +358,11 @@ namespace Microsoft.AppCenter.Crashes
                     attachment.ErrorId = errorId;
                     try
                     {
-                        attachment.ValidatePropertiesForAttachment();
+                        if (!attachment.ValidatePropertiesForAttachment())
+                        {
+                            AppCenterLog.Error(LogTag, "Not all required fields are present in ErrorAttachmentLog.");
+                            continue;
+                        }
                         ++totalErrorAttachments;
                         tasks.Add(Channel.EnqueueAsync(attachment));
                     }
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
@@ -358,7 +358,7 @@ namespace Microsoft.AppCenter.Crashes
                     attachment.ErrorId = errorId;
                     try
                     {
-                        attachment.Validate();
+                        attachment.ValidatePropertiesForAttachment();
                         ++totalErrorAttachments;
                         tasks.Add(Channel.EnqueueAsync(attachment));
                     }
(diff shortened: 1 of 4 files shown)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -358,7 +358,11 @@ private Task SendErrorAttachmentsAsync(Guid errorId, IEnumerable<ErrorAttachment
                     attachment.ErrorId = errorId;
                     try
```
MatkovIvan: It can be removed now

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
```diff
@@ -89,5 +89,14 @@ public override void Validate()
                 throw new ValidationException(ValidationException.Rule.CannotBeNull, "Data");
             }
         }
+
+        public bool ValidatePropertiesForAttachment()
```
jaelim-ms: Should we do this in `Validate`?

Tests/Microsoft.AppCenter.Crashes.Test.Windows/ErrorAttachmentLogTest.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
```
jaelim-ms: We've talked about this one offline. We should use the default format configuration that won't bring `System` usings to the top.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -356,16 +356,13 @@ private Task SendErrorAttachmentsAsync(Guid errorId, IEnumerable<ErrorAttachment
                 {
                     attachment.Id = Guid.NewGuid();
                     attachment.ErrorId = errorId;
-                    try
+                    if (!attachment.ValidatePropertiesForAttachment())
```
achocron: Why not use the Validate method (and just modify it?) Is it because it is missing something like device at this point? If that's the reason, can you just add a comment explaining?

achocron: Also, maybe invert the `if`

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/ErrorAttachmentLog.cs
achocron: I think Validate will do too much validation for the point at which this method gets called.

achocron: But I agree maybe it should be called in validate?

MatkovIvan: @jaelim-ms Nope. It was the problem.

We have similar way in iOS (separate validate function)

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
MatkovIvan: Yes, the device will be initialized later by channel.
We have similar way in iOS (separate validate function).

Let's keep validate **base** log class as is and just add a comment here

guperrot: Also the code is generated and swagger spec is not good enough for that for some models.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/ErrorAttachmentLogTest.cs
guperrot: not a blocker though

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
guperrot: @AnastasiaKubova can you address https://github.com/microsoft/appcenter-sdk-dotnet/pull/1021/files#r301215312 in another PR? I unblocked this one as good enough to unblock test app review.

AnastasiaKubova: Ok

PR #1023, opened 2019-07-08
Add  crashes callbacks functionality to test apps

AnastasiaKubova:

* [ ] ~Are tests passing locally?~
* [ ] ~Did you add unit tests if this modifies the UWP implementation?~

## Description

Add functionality to the WPF app for the following APIs: Events/callbacks, HasCrashedInLastSession, GetLastSessionErrorReportAsync, and user confirmation.

image

## Related PRs or issues

[AB#63633](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/63633)

### DIFF
--- a/Apps/Contoso.WPF.Demo/App.xaml.cs
+++ b/Apps/Contoso.WPF.Demo/App.xaml.cs
@@ -14,7 +14,7 @@ namespace Contoso.WPF.Demo
     /// </summary>
     public partial class App
     {
-        private const string LogTag = "AppCenterSasquatch";
+        private const string LogTag = "AppCenterDemo";
 
         protected override void OnStartup(StartupEventArgs e)
         {
@@ -22,7 +22,7 @@ namespace Contoso.WPF.Demo
             Crashes.ShouldAwaitUserConfirmation = ConfirmationHandler;
             Crashes.ShouldProcessErrorReport = ShouldProcess;
 
-            // Event handlers
+            // Event handlers.
             Crashes.SendingErrorReport += SendingErrorReportHandler;
             Crashes.SentErrorReport += SentErrorReportHandler;
             Crashes.FailedToSendErrorReport += FailedToSendErrorReportHandler;
@@ -60,8 +60,6 @@ namespace Contoso.WPF.Demo
         {
             AppCenterLog.Info(LogTag, "Sending error report");
             var report = e.Report;
-
-            // Test some values
             if (report.Exception != null)
             {
                 AppCenterLog.Info(LogTag, report.Exception.ToString());
@@ -72,8 +70,6 @@ namespace Contoso.WPF.Demo
         {
             AppCenterLog.Info(LogTag, "Sent error report");
             var report = e.Report;
-
-            // Test some values
             if (report.Exception != null)
             {
                 AppCenterLog.Info(LogTag, report.Exception.ToString());
@@ -88,8 +84,6 @@ namespace Contoso.WPF.Demo
         {
             AppCenterLog.Info(LogTag, "Failed to send error report");
             var report = e.Report;
-
-            // Test some values
             if (report.Exception != null)
             {
                 AppCenterLog.Info(LogTag, report.Exception.ToString());
--- a/Apps/Contoso.WPF.Demo/App.xaml.cs
+++ b/Apps/Contoso.WPF.Demo/App.xaml.cs
@@ -1,11 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+using System.Globalization;
+using System.Windows;
 using Microsoft.AppCenter;
 using Microsoft.AppCenter.Analytics;
 using Microsoft.AppCenter.Crashes;
-using System;
-using System.Windows;
 
 namespace Contoso.WPF.Demo
 {
@@ -14,8 +15,6 @@ namespace Contoso.WPF.Demo
     /// </summary>
     public partial class App
     {
-        private const string LogTag = "AppCenterDemo";
-
         protected override void OnStartup(StartupEventArgs e)
         {
             AppCenter.LogLevel = LogLevel.Verbose;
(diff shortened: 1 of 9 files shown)


Apps/Contoso.WPF.Demo/MainWindow.xaml.cs
```diff
@@ -43,6 +43,7 @@ public MainWindow()
         private void UpdateState()
         {
             AppCenterEnabled.IsChecked = AppCenter.IsEnabledAsync().Result;
+            CrashesEnabled.IsChecked = Crashes.IsEnabledAsync().Result;
```
MatkovIvan: Good catch :+1:

Apps/Contoso.WPF.Demo/App.xaml.cs
```diff
@@ -13,10 +14,84 @@ namespace Contoso.WPF.Demo
...
+            Crashes.ShouldProcessErrorReport = ShouldProcess;
+
+            // Event handlers.
+            Crashes.SendingErrorReport += SendingErrorReportHandler;
+            Crashes.SentErrorReport += SentErrorReportHandler;
+            Crashes.FailedToSendErrorReport += FailedToSendErrorReportHandler;
             AppCenter.Start("f4e2a83d-3052-4884-8176-8b2c50277d16", typeof(Analytics), typeof(Crashes));
+            Crashes.HasCrashedInLastSessionAsync().ContinueWith(hasCrashed =>
+            {
+                AppCenterLog.Info(LogTag, "Crashes.HasCrashedInLastSession=" + hasCrashed.Result);
```
jaelim-ms: Please do not use `AppCenterLog` in Demo and Test app. This should use its own logging framework.

Apps/Contoso.WPF.Demo/App.xaml.cs
```diff
@@ -13,10 +14,84 @@ namespace Contoso.WPF.Demo
...
+            Crashes.GetLastSessionCrashReportAsync().ContinueWith(task =>
+            {
+                AppCenterLog.Info(LogTag, "Crashes.LastSessionCrashReport.Exception=" + task.Result?.Exception);
+            });
+        }
+
+        bool ShouldProcess(ErrorReport report)
+        {
+            AppCenterLog.Info(LogTag, "Determining whether to process error report");
+            return true;
```
jaelim-ms: Better to have a UI to set this value for testing. Add a UI component and have either `true` or `false` that could be selected by testers, persist it in the application and read the value in the method.
I'm not sure this isn't in scope but this will allow us to test ShouldProcess that returns both `true` and `false`.

Apps/Contoso.WPF.Demo/App.xaml.cs
```diff
@@ -13,10 +14,84 @@ namespace Contoso.WPF.Demo
     /// </summary>
     public partial class App
     {
+        private const string LogTag = "AppCenterDemo";
+
         protected override void OnStartup(StartupEventArgs e)
         {
             AppCenter.LogLevel = LogLevel.Verbose;
+            Crashes.ShouldAwaitUserConfirmation = ConfirmationHandler;
```
jaelim-ms: Simply return `true` unless you want to address my feedback below.

Apps/Contoso.WPF.Demo/App.xaml.cs
```diff
@@ -13,10 +14,84 @@ namespace Contoso.WPF.Demo
...
+            return true;
+        }
+
+        static void SendingErrorReportHandler(object sender, SendingErrorReportEventArgs e)
+        {
+            AppCenterLog.Info(LogTag, "Sending error report");
+            var report = e.Report;
+            if (report.Exception != null)
+            {
+                AppCenterLog.Info(LogTag, report.Exception.ToString());
```
jaelim-ms: Doesn't seem `report.Exception.ToString()` is useful. Please log at least error ID.

Apps/Contoso.WPF.Puppet/App.xaml.cs
```diff
@@ -14,14 +14,85 @@ namespace Contoso.WPF.Puppet
...
+        private const string LogTag = "AppCenterPuppet";
+
         protected override void OnStartup(StartupEventArgs e)
         {
             AppCenter.LogLevel = LogLevel.Verbose;
             AppCenter.SetLogUrl("https://in-integration.dev.avalanch.es");
-            Crashes.SendingErrorReport += (sender, args) => Console.WriteLine($@"[App] Sending report Id={args.Report.Id} Exception={args.Report.Exception}");
-            Crashes.SentErrorReport += (sender, args) => Console.WriteLine($@"[App] Sent report Id={args.Report.Id}");
-            Crashes.FailedToSendErrorReport += (sender, args) => Console.WriteLine($@"[App] FailedToSend report Id={args.Report.Id} Error={args.Exception}");
+            Crashes.ShouldAwaitUserConfirmation = ConfirmationHandler;
```
jaelim-ms: Same comments as Demo app. Please address all comments in Demo app here.

Apps/Contoso.WPF.Demo/App.xaml.cs
jaelim-ms: This is aligning with other platforms but probably a good to have one for testing in the future.

Apps/Contoso.WPF.Demo/App.xaml.cs
```diff
@@ -16,7 +18,49 @@ public partial class App
...
+            });
+            Crashes.GetLastSessionCrashReportAsync().ContinueWith(task =>
+            {
+                Log("Crashes.LastSessionCrashReport.Exception=" + task.Result?.Exception);
+            });
+        }
+
+        private static bool ConfirmationHandler()
+        {
+            Current.Dispatcher.Invoke(() =>
```
guperrot: This is blocking until we clicked, it works but it makes logging odd and the intent is to be "asynchronous" here.

guperrot: #1024

PR #1027, opened 2019-07-09
[Fix] Both forms app send 1 push log for every character typed in userId field

russelarms:

## Description

Use either Contoso.Forms.Puppet or Contoso.Forms.Demo on Android or iOS.
Type characters in userId.

Expected: push log sent only when done typing.
Actual: 1 push log sent for every typed character.

## Related PRs or issues

[AB#64145](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/64145)

### DIFF
--- a/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/Constants.cs
+++ b/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/Constants.cs
@@ -1,0 +1,9 @@
+﻿using System;
+namespace Contoso.Forms.Demo
+{
+    [Android.Runtime.Preserve(AllMembers = true)]
+    public static class Constants
+    {
+        public const string UserId = "userId";
+    }
+}
--- a/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/Constants.cs
+++ b/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/Constants.cs
@@ -1,8 +1,5 @@
-﻿using System;
-
-namespace Contoso.Forms.Demo
+﻿namespace Contoso.Forms.Demo
 {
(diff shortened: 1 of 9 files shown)


russelarms:
I removed bindings as EntryCell has no Unfocus event to subscribe.

Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/Constants.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+namespace Contoso.Forms.Demo
```
MatkovIvan:
```suggestion

namespace Contoso.Forms.Demo
```

Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/Constants.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+
+namespace Contoso.Forms.Demo
+{
+    [Android.Runtime.Preserve(AllMembers = true)]
```
guperrot: Was the symbol removed in release?, it's suprising as there is no reflection here so should not be necessary.

Did you actually test in release and had a crash when this attribute was missing on this constant?

MatkovIvan: @guperrot I guess it's just to make the things the same as here
https://github.com/microsoft/appcenter-sdk-dotnet/pull/1027/files#diff-3be604fc8d5e709029be52313f8b20b3R6

guperrot: Let's not introduce the annotation if not necessary, please run locally in release and verify.

PR #1028, opened 2019-07-09
Fix/clear managed error logs

michaelsmitha:

## Description

Should check for IsInactive before running HandleUserConfirmation code

## Related PRs or issues

[AB#64333](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/64333)

### DIFF
--- a/Tests/Microsoft.AppCenter.Crashes.Test.Windows/CrashesTest.cs
+++ b/Tests/Microsoft.AppCenter.Crashes.Test.Windows/CrashesTest.cs
@@ -551,6 +551,24 @@ namespace Microsoft.AppCenter.Crashes.Test.Windows
         }
 
         [TestMethod]
+        public void DisablingCrashesCleansUpUnprocessedManagedErrorLogs()
+        {
+            // Make a pending error log file.
+            ErrorLogHelper.Instance = GenerateMockErrorLogHelperWithPendingFile();
+
+            // Start Crashes.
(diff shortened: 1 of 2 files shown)


Tests/Microsoft.AppCenter.Crashes.Test.Windows/CrashesTest.cs
```diff
@@ -776,6 +794,22 @@ public void HandleUserConfirmationAsyncAlwaysSend()
...
+
+            // Start Crashes.
+            Crashes.SetEnabledAsync(false).Wait();
+
+            Crashes.Instance.OnChannelGroupReady(_mockChannelGroup.Object, string.Empty);
+            Crashes.NotifyUserConfirmation(UserConfirmation.AlwaysSend);
+
+            // The alwaysSend value should never set because we are disabled
+            var alwaysSendValue = AppCenter.Instance.ApplicationSettings.GetValue(Crashes.PrefKeyAlwaysSend, false);
+            Assert.IsFalse(alwaysSendValue);
```
jaelim-ms: Is there anything else we can verify other than always send flag?

guperrot: no as the collection is empty as its cleaned, so foreach loops are empty.
We want to see that it prints log of being inactive but we don't have asserts for logging.

PR #1029, opened 2019-07-09
Implement handled errors for windows

guperrot:

* [x] Did you add unit tests?

## Description

Implement handled errors for windows.
This can be tested using contoso.forms.puppet.uwp right now.

[AB#64394](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/64394)


SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Analytics.cs
```diff
@@ -9,17 +9,15 @@
 using Microsoft.AppCenter.Ingestion.Models.Serialization;
 using Microsoft.AppCenter.Utils;
 using System.Threading.Tasks;
+using Microsoft.AppCenter.Windows.Shared.Utils;
```
jaelim-ms: Order?

guperrot: e52e93e

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Ingestion/Models/EventLog.cs
```diff
@@ -80,16 +84,15 @@ public override void Validate()
             base.Validate();
             if (Name == null)
             {
-                throw new ValidationException(ValidationException.Rule.CannotBeNull, nameof(Name));
+                throw new ValidationException(ValidationException.Rule.CannotBeNull, "Name");
```
clpolet: Worth a constant?

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Ingestion/Models/PageLog.cs
```diff
@@ -41,14 +39,23 @@ public partial class PageLog : LogWithProperties
         /// Concrete types like StartSessionLog or PageLog are always part of a
         /// session and always include this identifier.
         /// </param>
+        /// <param name="userId">optional string used for associating logs with
```
clpolet:
```suggestion
        /// <param name="userId">Optional string used for associating logs with
```

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Ingestion/Models/EventLog.cs
```diff
@@ -46,15 +41,24 @@ public partial class EventLog : LogWithProperties
         /// Concrete types like StartSessionLog or PageLog are always part of a
         /// session and always include this identifier.
         /// </param>
+        /// <param name="userId">optional string used for associating logs with
```
clpolet:
```suggestion
        /// <param name="userId">Optional string used for associating logs with
```

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Ingestion/Models/LogWithProperties.cs
```diff
@@ -33,14 +30,23 @@ public abstract class LogWithProperties : Log
         /// Concrete types like StartSessionLog or PageLog are always part of a
         /// session and always include this identifier.
         /// </param>
+        /// <param name="userId">optional string used for associating logs with
```
clpolet:
```suggestion
        /// <param name="userId">Optional string used for associating logs with
```

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Ingestion/Models/EventLog.cs
guperrot: This is generated code

guperrot:
@clpolet all the models are from autorest generated code, we remove only things that don't compile due to dependencies.

jaelim-ms:
@clpolet This could be fixed by changing yaml file [here](https://msmobilecenter.visualstudio.com/Mobile-Center/_git/appcenter?path=%2Fingestion%2Fapi-specs%2Fingestion.yaml&version=GBmaster&line=92&lineStyle=plain&lineEnd=93&lineStartColumn=1&lineEndColumn=1)

guperrot:
> @clpolet This could be fixed by changing yaml file [here](https://msmobilecenter.visualstudio.com/Mobile-Center/_git/appcenter?path=%2Fingestion%2Fapi-specs%2Fingestion.yaml&version=GBmaster&line=92&lineStyle=plain&lineEnd=93&lineStartColumn=1&lineEndColumn=1)

👍  for fixing source of truth then regenerating as opposed to tweaking

Tests/Microsoft.AppCenter.Crashes.Test.Windows/CrashesTest.cs
```diff
@@ -810,6 +817,39 @@ public void HandleUserConfirmationAsyncAlwaysSendWhileDisabled()
...
+                    actualLog = JsonConvert.DeserializeObject<LogContainer>(content, LogSerializer.SerializationSettings).Logs.Single() as HandledErrorLog;
+                    semaphore.Release();
+                });
+
+            var exception = new System.Exception("Something went wrong.");
+            var properties = new Dictionary<string, string> { { "k1", "v1" }, { "p2", "v2"} };
+            Crashes.TrackError(exception, properties);
+
+            // Wait until the http layer sends the log.
+            semaphore.Wait();
```
guperrot: In case the test fails, it will hang.
Though not a big fan of timeouts (since its like sleeping in a way), I would put a time limit here.
Anyway wait is like sleep infinite, so finite is still better than infinite.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/PropertyValidator.cs
```diff
@@ -0,0 +1,67 @@
...
+        private const int MaxProperties = 20;
+        private const int MaxPropertyKeyLength = 125;
+        private const int MaxPropertyValueLength = 125;
+
+        /// <summary>
+        /// Validates properties.
+        /// </summary>
+        /// <param name="properties">Properties collection to validate.</param>
+        /// <param name="logName">Log name.</param>
+        /// <returns>Valid properties collection with maximum size of 5</returns>
```
michaelsmitha: `<returns>Valid properties collection with maximum size of 5</returns>`

Is this true? Or should it be `with maximum size of {MaxProperties}`?

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Ingestion/Models/EventLog.cs
```diff
@@ -46,15 +41,24 @@ public partial class EventLog : LogWithProperties
...
+        public EventLog(Device device, System.Guid id, string name, System.DateTime? timestamp = default(System.DateTime?), System.Guid? sid = default(System.Guid?), string userId = default(string), IDictionary<string, string> properties = default(IDictionary<string, string>))
+            : base(device, timestamp, sid, userId, properties)
         {
             Id = id;
             Name = name;
+            CustomInit();
         }
 
+        /// <summary>
+        /// An initialization method that performs custom operations like setting defaults
```
michaelsmitha:
```suggestion
        /// An initialization method that performs custom operations like setting defaults.
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
```diff
@@ -0,0 +1,83 @@
...
+
+namespace Microsoft.AppCenter.Crashes.Ingestion.Models
+{
+    using Microsoft.AppCenter.Ingestion.Models;
+    using Newtonsoft.Json;
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Handled Error log for managed platforms (such as Xamarin, Unity,
+    /// Android Dalvik/ART)
```
michaelsmitha:
```suggestion
    /// Android Dalvik/ART).
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
```diff
@@ -0,0 +1,83 @@
...
+        public HandledErrorLog(Device device, Exception exception, System.DateTime? timestamp = default(System.DateTime?), System.Guid? sid = default(System.Guid?), string userId = default(string), IDictionary<string, string> properties = default(IDictionary<string, string>), System.Guid? id = default(System.Guid?))
+            : base(device, timestamp, sid, userId, properties)
+        {
+            Id = id;
+            Exception = exception;
+            CustomInit();
+        }
+
+        /// <summary>
+        /// An initialization method that performs custom operations like setting defaults
```
michaelsmitha:
```suggestion
        /// An initialization method that performs custom operations like setting defaults.
```

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
```diff
@@ -0,0 +1,83 @@
...
+        partial void CustomInit();
+
+        /// <summary>
+        /// Gets or sets unique identifier for this Error.
+        ///
```
michaelsmitha:
```suggestion
```

Tests/Microsoft.AppCenter.Crashes.Test.Windows/CrashesTest.cs
```diff
@@ -4,13 +4,20 @@
 using Microsoft.AppCenter.Channel;
 using Microsoft.AppCenter.Crashes.Ingestion.Models;
 using Microsoft.AppCenter.Crashes.Utils;
+using Microsoft.AppCenter.Ingestion.Http;
```
jaelim-ms: Was the test broken? I wonder why it has new `using`s without adding any code.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// </summary>
+        [JsonProperty(PropertyName = "id")]
+        public System.Guid? Id { get; set; }
+
+        /// <summary>
+        /// </summary>
```
michaelsmitha: Empty  Summary

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
```diff
@@ -0,0 +1,83 @@
...
+        [JsonProperty(PropertyName = "exception")]
+        public Exception Exception { get; set; }
+
+        /// <summary>
+        /// Validate the object.
+        /// </summary>
+        /// <exception cref="ValidationException">
+        /// Thrown if validation fails
```
michaelsmitha:
```suggestion
        /// Thrown if validation fails.
```

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+                {
+                    actualLog = JsonConvert.DeserializeObject<LogContainer>(content, LogSerializer.SerializationSettings).Logs.Single() as HandledErrorLog;
+                    semaphore.Release();
+                });
+            var exception = new System.Exception("Something went wrong.");
+            var properties = new Dictionary<string, string> { { "k1", "v1" }, { "p2", "v2" } };
+            Crashes.TrackError(exception, properties);
+
+            // Wait until the http layer sends the log.
+            semaphore.Wait(1000);
```
jaelim-ms: Can we refactor an internal method that returns a Task (or assigns a Task to a property)? This is going to be flaky at some points. I also thought about this when I implemented `ProcessPendingErrorsAsync`. `ProcessPendingErrorsTask` property is the one that I added for testing purpose.
Or, is this something that we can resolve in functional test? Then I'm fine.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+            Assert.IsNull(actualLog.Properties);
+        }
+
+        [TestMethod]
+        public void TrackErrorWithoutException()
+        {
+            var mockHttpNetworkAdapter = StartCrashes();
+            var semaphore = new SemaphoreSlim(0);
+            HandledErrorLog actualLog = null;
+            Mock.Get(mockHttpNetworkAdapter).Setup(adapter => adapter.SendAsync(
```
jaelim-ms: The test scenario doesn't require this mock unless you fail the test in the callback.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace Microsoft.AppCenter.Crashes.Test.Windows
+{
+    [TestClass]
+    public class HandledErrorTest
+    {
+        private static IHttpNetworkAdapter StartCrashes()
```
jaelim-ms: Can't it be `TestInitialize`? The method is called at the beginning of all tests.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Ingestion/Models/LogWithProperties.cs
```diff
@@ -33,14 +30,23 @@ public abstract class LogWithProperties : Log
...
-            : base(device, timestamp, sid)
+        public LogWithProperties(Device device, System.DateTime? timestamp = default(System.DateTime?), System.Guid? sid = default(System.Guid?), string userId = default(string), IDictionary<string, string> properties = default(IDictionary<string, string>))
+            : base(device, timestamp, sid, userId)
         {
             Properties = properties;
+            CustomInit();
         }
 
+        /// <summary>
+        /// An initialization method that performs custom operations like setting defaults
```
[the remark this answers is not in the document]
jaelim-ms: We don't want to change generated models. This should be changed in YAML.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
jaelim-ms: We don't want to change generated models. This should be changed in YAML.

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Ingestion/Models/EventLog.cs
jaelim-ms: We don't want to change generated models. This should be changed in YAML.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
jaelim-ms: We don't want to change generated models. This should be changed in YAML.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
jaelim-ms: We don't want to change generated models. This should be changed in YAML.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
jaelim-ms: We don't want to change generated models. This should be changed in YAML.

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Ingestion/Models/HandledErrorLog.cs
jaelim-ms: We don't want to change generated models. This should be changed in YAML.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+                    It.IsAny<CancellationToken>()))
+                .Callback((string uri, string method, IDictionary<string, string> headers, string content, CancellationToken cancellation) =>
+                {
+                    actualLog = JsonConvert.DeserializeObject<LogContainer>(content, LogSerializer.SerializationSettings).Logs.Single() as HandledErrorLog;
+                    semaphore.Release();
+                });
+            Crashes.TrackError(null);
+
+            // Check no log sent.
+            semaphore.Wait(1000);
```
jwallra: Why do we have to `Wait()` for this test?

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+            return mockHttpNetworkAdapter;
+        }
+
+        [TestMethod]
+        public void TrackErrorWithValidParameters()
+        {
+            var mockHttpNetworkAdapter = StartCrashes();
+            var semaphore = new SemaphoreSlim(0);
+            HandledErrorLog actualLog = null;
+            Mock.Get(mockHttpNetworkAdapter).Setup(adapter => adapter.SendAsync(
```
jwallra: This is the exact same for several tests, can we do the setup in the initialization or move it to a helper method?

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+                It.IsAny<IDictionary<string, string>>(),
+                It.IsAny<string>(),
+                It.IsAny<CancellationToken>()))
+                .Callback((string uri, string method, IDictionary<string, string> headers, string content, CancellationToken cancellation) =>
+                {
+                    actualLog = JsonConvert.DeserializeObject<LogContainer>(content, LogSerializer.SerializationSettings).Logs.Single() as HandledErrorLog;
+                    semaphore.Release();
+                });
+            var exception = new System.Exception("Something went wrong.");
+            var maxProperties = 20;
```
jwallra: I'm not a fan of this pattern because it requires us to modify the tests if we change the number of max properties.  The test validation doesn't change though, so it's just making more work for us later by not using the constants.

I would prefer to make the constants `internal` instead of re-defining them

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
achocron: I'm not sure the solution will work here. `Crashes` can get a `Task` for the channel's `EnqueueAsync` and expose it to the test, but that task returns once the log is stored to disk. It doesn't wait until the crash was sent to ingestion. Thus, we would need to significantly refactor the `Channel`. So there is no really good way to avoid the timeout.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
achocron: Can you elaborate? Seems like it is necessary to capture the log that is enqueued.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+                    actualLog = JsonConvert.DeserializeObject<LogContainer>(content, LogSerializer.SerializationSettings).Logs.Single() as HandledErrorLog;
+                    semaphore.Release();
+                });
+            var exception = new System.Exception("Something went wrong.");
+            var properties = new Dictionary<string, string>
+            {
+                [""] = "testEmptyKeySkipped",
+                ["testNullValueSkipped"] = null,
+                [new string('k', 126)] = "testKeyTooLong",
+                ["testValueTooLong"] = new string('v', 126),
```
jwallra: Please make constants internal and use them in the tests - see reasoning above.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+                {
+                    actualLog = JsonConvert.DeserializeObject<LogContainer>(content, LogSerializer.SerializationSettings).Logs.Single() as HandledErrorLog;
+                    semaphore.Release();
+                });
+            var exception = new System.Exception("Something went wrong.");
+            var properties = new Dictionary<string, string>
+            {
+                [""] = "testEmptyKeySkipped",
+                ["testNullValueSkipped"] = null,
+                [new string('k', 126)] = "testKeyTooLong",
```
jwallra: Please make constants internal and use them in the tests - see reasoning above.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+            Assert.AreEqual(exception.Message, actualLog.Exception.Message);
+
+            // Check original dictionary not sent (copy is made).
+            Assert.IsNotNull(actualLog.Properties);
+            Assert.AreNotSame(properties, actualLog.Properties);
+
+            // Check invalid values were skipped or truncated.
+            Assert.AreEqual(3, actualLog.Properties.Count);
+            Assert.AreEqual("normalValue", actualLog.Properties["normalKey"]);
+            Assert.AreEqual("testKeyTooLong", actualLog.Properties[new string('k', 125)]);
```
jwallra: Please make constants internal and use them in the tests - see reasoning above.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
```diff
@@ -0,0 +1,237 @@
...
+            Assert.AreEqual(new string('v', 125), actualLog.Properties["testValueTooLong"]);
```
jwallra: Please make constants internal and use them in the tests - see reasoning above.

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Ingestion/Models/EventLog.cs
achocron: Generated code.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
jaelim-ms: Then is 1 sec wait enough? Channel is configured to send logs every 1 sec if there are logs. It should be waiting more than 1 sec but I don't think it is worth to address it here or discuss more about this one since we have a plan to change it already. 😄

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
jaelim-ms: There are pros and cons but I personally also don't like this one. Not a big deal though.

achocron: I think that if we are relying on the constants we have defined in the code, then the test is quite circular. So I actually prefer redefinition in tests.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
achocron: I personally think that the test gets a bit obfuscated the more we move into the setup. Especially with mocking

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
achocron: I think that it just gives a chance for the test to fail

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
jaelim-ms: Is `SendAsync` really called for `null` exception?

achocron: No, but that is what this test is verifying

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
achocron: Since this is more of an opinion type thing, I'm going to resolve the comments

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
jaelim-ms: Assert fail in the callback would be a good option to resolve this comment.

achocron: Done.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
achocron: Resolving as this is also a bit more of an opinion type thing.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
jwallra: This is just less maintainable than making the constants internal, though.  I understand that by a "strict" definition it's circular, but I'm not sure that's a good reason to use this pattern.

If someone looks at this code and sees the number 20, they'll wonder where it's coming from.  There's no logical link between this value and the one in the test file, you just have to look in the file being tested.  What if the values aren't even in that file?  It becomes very difficult to link the verification value to the "source of truth"

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/PropertyValidator.cs
guperrot: We need to update swagger then

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
guperrot: We can't, try it ;) You will get an error that an out variable cannot be set from a lambda expression.

Tests/Microsoft.AppCenter.Crashes.Test.Windows/HandledErrorTest.cs
guperrot: A test should not use constants from the SDK, it's a contract.
We already decided in GDPR test runner to not use the models, the same reasoning applies here.

achocron: Found a compromise: We use the constants in tests for maintainability, but in order to make sure they still have the right values, we add specific tests for that elsewhere. Thus, we reduce the magic values to only one test, but we are still testing that they are correct.

jwallra: More context: https://stackoverflow.com/questions/3360074/do-you-use-constants-from-the-implementation-in-your-test-cases/23374335#23374335

PR #1032, opened 2019-07-10
Crash logs a warning with exception when starting in disabled mode

michaelsmitha:

## Description

On startup with Crashes disabled or toggling off Crashes without a crashes directory would WARN with a NotFoundException.

## Related PRs or issues

[AB#64338](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/64338)

## Misc

I also found that toggling the module on and off would throw this warning with every off-toggle even though the directory hadn't been recreated.
Without the `.Refresh()` call, the `.Exists()` call would remain `true` post-delete.

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
@@ -305,19 +305,23 @@ namespace Microsoft.AppCenter.Crashes.Utils
 
         public virtual void InstanceRemoveAllStoredErrorLogFiles()
         {
-            lock (LockObject)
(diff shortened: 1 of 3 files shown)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Utils/ErrorLogHelper.cs
```diff
@@ -307,17 +307,23 @@ public virtual void InstanceRemoveAllStoredErrorLogFiles()
         {
             lock (LockObject)
             {
-                AppCenterLog.Debug(Crashes.LogTag, $"Deleting error log directory.");
-                try
-                {
-                    _crashesDirectory.Delete(true);
-                }
-                catch (System.Exception ex)
+                _crashesDirectory.Refresh();
```
guperrot: if just calling Exists without Refresh it will fail?

PR #1036, opened 2019-07-12
Handle sessionId in crash correctly for breadcrumbs

guperrot:

## Description

Handle sessionId in crash correctly for breadcrumbs

## Related PRs or issues

[AB#64337](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/64337)


SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Channel/SessionTracker.cs
```diff
@@ -86,6 +86,14 @@ public void Resume()
             }
         }
 
+        public void Stop()
+        {
+            lock (_lockObject)
+            {
+                UpdateSessionId(Guid.Empty);
```
guperrot: Test missing for that one.
Also can we get rid of LastSid now that we have this mechanism? Several tests depend on it so not easy to refactor.

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Channel/SessionTracker.cs
```diff
@@ -96,26 +100,29 @@ private void HandleEnqueuingLog(object sender, EnqueuingLogEventArgs e)
                 {
                     return;
                 }
-                e.Log.Sid = _sid;
+
+                // Correlate current session only to logs not specifying timestamps.
+                // Crash before restart when Analytics was disabled at that time must not have session identifier.
```
jaelim-ms: Can you explain how it knows it crashed with **Analytics disabled** by `timestamp == null`?
I can see the point that Crashes logs don't have `Timestamp` but what's the correlation between the module being disabled and timestamp?

Tests/Microsoft.AppCenter.Analytics.Test.Windows/SessionTrackerTest.cs
```diff
@@ -97,22 +129,33 @@ public void ResumeAfterShortPause()
...
         }
 
+        /// <summary>
+        ///     Verify that an enqueuing log with already specified timestamp is ignored from session tracker, like crash logs.
+        /// </summary>
+        [TestMethod]
+        public void HandleEnqueuingLogWithExplicitTimestamp()
+        {
+            _sessionTracker.Resume();
+            var eventLog = new EventLog { Timestamp = DateTime.Now };
```
jaelim-ms: I could see the point of the test but it might mislead. Can't you use anonymous type of log or custom type of log here?
Not a big deal though.

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Windows.Shared/Channel/SessionTracker.cs
guperrot: When Analytics is disabled, the sessionId is missing from crash (analytics set sessionId to null on the context).
Since crash specified its own timestamp, we skip sessionId correlation.
Thus the sessionId will be null as expected.
This if check is in Android too except it uses session history in that case, which we don't need for windows as we don't have c/c++ crashes.

Tests/Microsoft.AppCenter.Analytics.Test.Windows/SessionTrackerTest.cs
guperrot: We should use the real log type here, we send only events from Analytics.

Issue #1104, opened 2019-08-30
WinForms apps are prevented from crashing


danAtAllworx:
### **Description**

When AppCenter is installed into a WinForms application, an unhandled exception that would have previously caused a crash will no longer crash the application. This can leave the application running in an unstable state. This is especially dangerous because there is no way for the application code to know that this has happened (see #1103) so it can't intelligently respond to the situation.

### **Repro Steps**

Using this code:

    internal class Form1 : Form
    {
        [STAThread]
        static void Main()
        {
            AppCenter.Start("id", typeof(Crashes));

            Application.EnableVisualStyles();
            Application.SetCompatibleTextRenderingDefault(false);
            Application.Run(new Form1());
        }

        public Form1()
        {
            var button = new Button {Text = @"CRASH!"};
            button.Click += (sender, args) => throw new NotImplementedException();
            Controls.Add(button);
        }
    }

If the `AppCenter.Start()` is commented out, then clicking the button will cause the application to crash. With the `AppCenter.Start()` there, clicking the button seems to do nothing. The application continues to run normally.

Even more confusing is that (because of the way AppCenter works) no crash report is sent until the application is manually restarted. This could result in multiple crash reports being queued up for different crashes over an arbitrarily long period of time (days??) before the application is finally restarted, at which point there will be multiple days worth of "crash" data waiting to be sent.

### **Details**

Platform: WinForms, .NET Framework 4.5.1
SDK Version: 2.3.0
OS: Windows 10

russelarms:
Hi @danAtAllworx! Thanks for reporting that, we'll look into it.

achocron:
Hey @danAtAllworx. Thank you again for bringing this up; we've been able to reproduce the issue and are working on a fix.

For your second question about crash reports not being sent until restart, this is indeed the intended behavior. In general, if the application is in a corrupted state, or crashing at all, the app won't necessarily be able to send reports properly, since the network code can take much longer than simply storing information to disk. Thus, the crashes are saved and then sent on startup.

guperrot:
Hi, we needed to change the behavior to avoid this issue.
Now our SDK will no longer subscribe to ThreadException as it has side effects on the application.
This is the documentation change that will be in the next release:

```
## Unhandled exceptions on WinForms applications

> [!NOTE]
> This section and following sub sections apply only to WinForms. You can skip this section if you are integrating the SDK on WPF.

[...]

Issue #1103, opened 2019-08-30
Need a way to add a custom exception handler


danAtAllworx:
### **Description**

When AppCenter is initialized, it either overrides or is overridden by any custom exception handler. Cross reference with [this SO question](https://stackoverflow.com/q/57717007/3119991). There does not appear to be any clean way to allow both AppCenter and a custom handler to both be executed in the event of an exception.

AppCenter should either provide a way for a custom exception handler to call it, or (probably better) should provide a callback hook for a custom exception handler.

### **Repro Steps**

Here is a simple example:

    internal class Form1 : Form
    {
        [STAThread]
        static void Main()
        {
            Application.ThreadException += (sender, args) =>
            {
                Console.WriteLine(@"Handler called");
                MessageBox.Show(@"Something bad happened. The application will now close.");
                Application.Exit();
            };
            AppCenter.Start("id", typeof(Crashes));

            Application.EnableVisualStyles();
            Application.SetCompatibleTextRenderingDefault(false);
            Application.Run(new Form1());
        }

        public Form1()
        {
            var button = new Button {Text = @"CRASH!"};
            button.Click += (sender, args) => throw new NotImplementedException();
            Controls.Add(button);
        }
    }

Note that in the code above, a custom handler is added before the call to `AppCenter.Start()`. This will result in only AppCenter being called when the button is clicked. The custom handler will not be called. If the custom handler is added after the call to `AppCenter.Start()` then only the custom handler will be called. AppCenter will not be called.

### **Details**

Platform: WinForms, .NET Framework 4.5.1
SDK Version: 2.3.0
OS: Windows 10

nanabite:
I agree - I have tried to use AppCenter in my application, but was not able to due to this behaviour. We use DevExpress and throw a ValidationException in cases where we want to surface validation issues to the user. Because this gets handled by AppCenter, the DevExpress framework is never made aware of the exception and doesn't show the validation.

Personally, I think there should be an option to completely disable the "crash" handling functionality and allow us to handle sending "Errors" and "Crashes" to AppCenter manually - I created an NLog target which works well for this purpose.

Jamminroot:
Hey, @nanabite, @danAtAllworx, thanks for bringing this to our attention, Ill let our PMs team look at your request.

danAtAllworx:
[...]

PR #1116, opened 2019-09-11
Unhandled exceptions: don't alter behavior of the app when adding our SDK

guperrot:
This reverts commit 8c061baaa18f49869e570f7cc91be4f4ce37e86c, reversing
changes made to c4cea8c0f2d0c7c74ae04e6500862388189a05dc.

## Description

On WinForms the fact that the app crashes or does not crash is controlled by the developer of the application.
Defining ThreadException handler actually alters the behavior of the application and still is not enough to let application crash anyway (that would need to also set unhandled exception mode).

We are going to avoid altering the application behavior when integrating our SDK by reverting the previous change on handlers and document instead how to let app crash or how to handle and track exceptions using TrackException.

## Related PRs or issues

[AB#67780](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/67780)
#1103
#1104

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 #### WinForms
 
-* **[Fix]** Don't prevent WinForms applications from crashing. If unhandled exceptions are handled by the application, they must now be reported using `Crashes.TrackError` to be displayed on AppCenter, see the public documentation for more details about this change.
+* **[Fix]** Don't prevent WinForms applications from crashing. If unhandled exceptions are handled by the application, they must now be reported using `Crashes.TrackError` to be displayed on AppCenter, see the [public documentation](https://docs.microsoft.com/en-us/appcenter/sdk/crashes/wpf-winforms) for more details about this change.
 
 ## Version 2.3.0-preview
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,17 @@
 
 ### App Center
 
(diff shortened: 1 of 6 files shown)


guperrot:
We will document how to force crash or how to handle/track exceptions.
We will also modify our test app to force crash to ease testing.

CHANGELOG.md
```diff
@@ -4,12 +4,18 @@
...
+
 * **[Fix]** Fix application version being reported for ClickOnce deployments.
 
 ### App Center Crashes
 
 * **[Breaking change]** Remove insecure implementation of the raw `ErrorReport.Exception` property (now always returns `null` and marked as obsolete), and provide `string StackTrace` property as an alternative on Xamarin, UWP, WPF and WinForms.
 
+#### WinForms
+
+* **[Fix]** Don't prevent WinForms applications from crashing. If unhandled exceptions are handled by the application, they must now be reported using `Crashes.TrackError` to be displayed on AppCenter, see the public documentation for more details about this change.
```
achocron:
```suggestion
* **[Fix]** Don't prevent WinForms applications from crashing. If unhandled exceptions are handled by the application, they must now be reported using `Crashes.TrackError` to be displayed on AppCenter, see the [public documentation](https://docs.microsoft.com/en-us/appcenter/sdk/crashes/wpf-winforms) for more details about this change.
```

PR #1120, opened 2019-09-23
Change application to library in push rdxml

achocron:

It's really dangerous to use the "application" tag in a library's rd xml, since an application my try to use the same tag and there can only be one. Libraries should use the "library" tag, since there can be several of those.


guperrot:
What is this about?

achocron:
@guperrot Found it while doing the UWP spike. Basically it fixes a warning that you see when you try to compile with an rd.xml file for the application - this warning is bad because it means that only one rd.xml will be actually used. To allow other rd.xml files to coexist, we must use the "library" tag

guperrot:
Is it shipped? Do we need changelog?

achocron:
I believe there is no customer impact since we don't ship it. Can impact apps that reference the project though (so just puppet apps).

PR #1112, opened 2019-09-10
Initial support for auth type selector in puppet test app

micarls:
* Add a new UI element which is a picker for authtype (currently B2C and AAD)
* Default to B2C
* Popup window telling user to restart app if auth type is changed
* Use different set of app secrets depending on auth type selected (guids for AAD empty for now, will fill in later)

PR #1135, opened 2019-10-01
Make demo app support AAD (similar to how test app already supports it)

micarls:

## Description

Test app already supports AAD. This makes demo app support it as well.

## Related PRs or issues

https://github.com/microsoft/appcenter-sdk-dotnet/pull/1112

## Misc

Tested the IOS and Android app secrets are working (via analytics)

### DIFF
--- a/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/App.xaml.cs
+++ b/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/App.xaml.cs
@@ -25,10 +25,21 @@ namespace Contoso.Forms.Demo
     {
         public const string LogTag = "AppCenterXamarinDemo";
 
-        // App Center keys
-        const string uwpKey = "5bce20c8-f00b-49ca-8580-7a49d5705d4c";
-        const string androidKey = "987b5941-4fac-4968-933e-98a7ff29237c";
-        const string iosKey = "fe2bf05d-f4f9-48a6-83d9-ea8033fbb644";
+        // App Center B2C secrets
+        static readonly IReadOnlyDictionary<string, string> B2CAuthAppSecrets = new Dictionary<string, string>
+        {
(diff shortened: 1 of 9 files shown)


guperrot:
Was it tested with at least analytics?

micarls:
I haven't tested with analytics yet. If that is the min bar, I will test with that.

guperrot:
That change is only to test app secret switching, so any way to verify it uses the right appSecret after restart is good, crashes works too.

micarls:
Makes sense, thanks.

micarls:
Yes, I can confirm both IOS and Android telemetry points are getting hit. I consider this ready to merge now.

Apps/Contoso.Forms.Demo/Contoso.Forms.Demo/App.xaml.cs
```diff
@@ -25,10 +26,21 @@ public partial class App
...
+            { XamarinDevice.Android, "987b5941-4fac-4968-933e-98a7ff29237c" },
+            { XamarinDevice.iOS, "fe2bf05d-f4f9-48a6-83d9-ea8033fbb644" }
+        };
+
+        // App Center AAD secrets
+        static readonly IReadOnlyDictionary<string, string> AADAuthAppSecrets = new Dictionary<string, string>
+        {
+            { XamarinDevice.UWP, "5bce20c8-f00b-49ca-8580-7a49d5705d4c" }, // same for all devices
+            { XamarinDevice.Android, "739eacee-42de-454c-b0d7-c093e765e009" },
+            { XamarinDevice.iOS, "e9f015c6-6c2b-4410-8053-70eaa52d90e0" }
```
guperrot: This is missing plist file modifications for distribute and auth

micarls: Good point, I've added it for both puppet (which was somehow missed) and demo. Still working on testing this new change.

PR #1147, opened 2019-10-08
Feature/update crash models

achocron:
Small PR to update crash models.


achocron:
Files in the folder `scripts/generated/Models` are the untouched autogenerated classes. Please do not request changes on those ones since they must remain the same as what was output by the tool.

PR #1148, opened 2019-10-08
Feature/uwp crash metadata refactor only

achocron:
This contains the initial round of refactoring required for UWP to augment crash logs. Currently, there is no behavior change. But don't review until the models PR has been merged and feature/uwp-crash-metadata updated.


michaelsmitha:
Instances of `saveWrapperException` and `loadWrapperException` need to save/load a string instead of utf-8 byte array

PR #1151, opened 2019-10-08
Send extra info for crash/error when .NET native is used

guperrot:
[AB#69733](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/69733)


SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.UWP/Utils/ErrorLogHelperPart.cs
```diff
@@ -27,18 +59,85 @@ internal static ErrorExceptionAndBinaries CreateModelExceptionAndBinaries(System
...
+                // When attaching a debugger in release, it will break into MissingRuntimeArtifactException, just click continue as it is actually caught and recovered by the lib.
+                var codeView = reader.ReadCodeViewDebugDirectoryData(codeViewEntry);
+                var pdbPath = Path.GetFileName(codeView.Path);
+                var endAddress = imageBase + reader.PEHeaders.PEHeader.SizeOfImage;
+                return new ModelBinary
+                {
+                    StartAddress = string.Format(CultureInfo.InvariantCulture, AddressFormat, imageBase.ToInt64()),
+                    EndAddress = string.Format(CultureInfo.InvariantCulture, AddressFormat, endAddress.ToInt64()),
+                    Path = pdbPath,
+                    Architecture = "unknown", // TODO remove when backend does not need this field anymore.
```
guperrot: AB#70806

Tests/Microsoft.AppCenter.Test.UWP/UWPCrashesTest.cs
```diff
@@ -9,10 +9,25 @@ namespace Microsoft.AppCenter.Test.UWP
...
         {
-            ManagedErrorLog managedErrorLog = ErrorLogHelper.CreateErrorLog(new System.Exception("Test Exception"));
-            Assert.AreEqual(managedErrorLog.AppLaunchTimestamp.Value.Kind, DateTimeKind.Utc);
+            try
+            {
+                throw new System.Exception("test exception");
+            }
+            catch (System.Exception e)
+            {
+                // TODO want to check binaries and frames, but this won't be available unless .net native is used.
```
jwallra: Why doesn't the guarded code work unless .NET native is used?  What happens now?

achocron: We can only populate the exception frames and binaries in .NET native - just a known limitation of the approach. So we can only validate this in release, which uses .NET native.

PR #1159, opened 2019-10-11
nuspec change for UWP symbolication

dhei:

## Related PRs or issues

[AB#69735](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/69735)

### DIFF
--- a/nuget/AppCenterCrashes.nuspec
+++ b/nuget/AppCenterCrashes.nuspec
@@ -43,7 +43,6 @@
     <!-- UWP -->
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.dll" target="lib/uap10.0" />
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.xml" target="lib/uap10.0" />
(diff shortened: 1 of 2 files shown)


nuget/AppCenterCrashes.nuspec
```diff
@@ -38,6 +43,7 @@
     <!-- UWP -->
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.dll" target="lib/uap10.0" />
     <file src="$uwp_dir$/Microsoft.AppCenter.Crashes.xml" target="lib/uap10.0" />
+    <file src="SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.UWP/Properties/Microsoft.AppCenter.Crashes.UWP.rd.xml" target="lib/uap10.0/Microsoft.AppCenter.Crashes/Properties/Microsoft.AppCenter.Crashes.rd.xml" />
```
guperrot: Remove anything related to the RD file and then we can merge that to the feature branch. Just confirmed it does not solve the exception we are seeing, and the exception is actually recovered (if no debugger attached it would not break into exceptions and everything works, no need to bump sdk version, no need for rd.xml file).

PR #1176, opened 2019-10-24
[1DS] Rewrite Start parsing function.

annakocheshkova:

## Description

[AB#71453](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/71453)

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
@@ -43,9 +43,14 @@ namespace Microsoft.AppCenter
 
             // Create a dictionary choosing the last secret value for each key. If the key has more than one secret value then select the last of them.
             var secretsDictionary = secretsGroup.ToDictionary(pair => pair.Key, pair => pair.Last().Last());
-            if (secretsDictionary.ContainsKey(TargetKeyName) || secretsDictionary.ContainsKey(AppSecretKeyName))
+            if (secretsDictionary.ContainsKey(TargetKeyName))
             {
-                AppCenterLog.Debug(AppCenterLog.LogTag, "Found named identifier in the secret; using as-is.");
+                AppCenterLog.Debug(AppCenterLog.LogTag, "Found 'target=' identifier in the secret; using as-is.");
+                return secrets;
+            }
+            if (secretsDictionary.ContainsKey(AppSecretKeyName))
+            {
+                AppCenterLog.Debug(AppCenterLog.LogTag, "Found 'appSecret=' identifier in the secret; using as-is.");
                 return secrets;
             }
             var platformSecret = string.Empty;
--- a/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
@@ -15,8 +15,8 @@ namespace Microsoft.AppCenter
     {
         const string SecretDelimiter = ";";
         const string PlatformKeyValueDelimiter = "=";
-        const string TargetPostfix = "Target";
-        const string SecretPostfix = "appSecret";
+        const string TargetPostfix = "target";
+        const string SecretPostfix = "appsecret";
 
         // Gets the first instance of an app sceret and/or target token corresponding to the given platform name, or returns the string 
         // as-is if no identifier can be found. Logs a message if no identifiers can be found.
@@ -43,43 +43,35 @@ namespace Microsoft.AppCenter
 
             // Create a dictionary choosing the last secret value for each key.
             var secretsDictionary = secretsGroup.ToDictionary(pair => pair.Key.Trim(), pair => pair.Last().Last().Trim(), StringComparer.OrdinalIgnoreCase);
-
-            var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
-            if (secretsDictionary.ContainsKey(TargetPostfix.ToLower()) || secretsDictionary.ContainsKey(SecretPostfix.ToLower()))
+            if (!secretsDictionary.ContainsKey(TargetPostfix) || !secretsDictionary.ContainsKey(SecretPostfix))
             {
                 AppCenterLog.Debug(AppCenterLog.LogTag, "Found named identifier in the secret; using as-is.");
                 return secrets;
             }
-
             var platformSecret = string.Empty;
             var platformTargetToken = string.Empty;
-            bool foundSecret = false;
-            bool foundTarget = false;
             if (secretsDictionary.ContainsKey(platformIdentifier))
             {
-                foundSecret = secretsDictionary.TryGetValue(platformIdentifier, out platformSecret);
+                secretsDictionary.TryGetValue(platformIdentifier, out platformSecret);
             }
             if (secretsDictionary.ContainsKey(platformTargetIdentifier))
             {
-                foundTarget = secretsDictionary.TryGetValue(platformTargetIdentifier, out platformTargetToken);
+                secretsDictionary.TryGetValue(platformTargetIdentifier, out platformTargetToken);
             }
-            var foundNoneOfTheKeys = !foundTarget && !foundSecret;
-            var bothKeysAreEmpty = string.IsNullOrEmpty(platformSecret) && string.IsNullOrEmpty(platformTargetToken);
-
-            if (foundNoneOfTheKeys || bothKeysAreEmpty)
+            if (string.IsNullOrEmpty(platformSecret) && string.IsNullOrEmpty(platformTargetToken))
             {
-                throw new AppCenterException(parseErrorMessage);
+                throw new AppCenterException($"Error parsing key for '{platformIdentifier}'");
             }
 
             // Format the string as "appSecret={};target={}" or "target={}" if needed.
-            if (platformTargetToken.Length > 0)
+            if (string.IsNullOrEmpty(platformTargetToken))
             {
-                //If there is an app secret
+                // If there is an app secret
                 if (platformSecret.Length > 0)
                 {
                     platformSecret = SecretPostfix + PlatformKeyValueDelimiter + platformSecret + SecretDelimiter;
                 }
-                platformSecret += TargetPostfix.ToLower() + PlatformKeyValueDelimiter + platformTargetToken;
+                platformSecret += TargetPostfix + PlatformKeyValueDelimiter + platformTargetToken;
             }
             return platformSecret;
         }
--- a/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
@@ -60,10 +60,10 @@ namespace Microsoft.AppCenter
             {
                 foundTarget = secretsDictionary.TryGetValue(platformTargetIdentifier, out platformTargetToken);
             }
-            var foundSomething = foundTarget || foundSecret;
-            var someKeyExists = !string.IsNullOrEmpty(platformSecret) || !string.IsNullOrEmpty(platformTargetToken);
+            var foundNoneOfTheKeys = !foundTarget && !foundSecret;
+            var bothKeysAreEmpty = string.IsNullOrEmpty(platformSecret) && string.IsNullOrEmpty(platformTargetToken);
 
-            if (!foundSomething || !someKeyExists)
+            if (foundNoneOfTheKeys || bothKeysAreEmpty)
             {
                 throw new AppCenterException(parseErrorMessage);
             }
--- a/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
@@ -38,7 +38,7 @@ namespace Microsoft.AppCenter
             var platformTargetIdicator = platformTargetIdentifier + "=";
             var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
             var targetTokenIdx = secrets.IndexOf(platformTargetIdicator, StringComparison.Ordinal);
-            var targetIdx = secrets.IndexOf(targetKeyword.toLowerCase(), StringComparison.Ordinal);
+            var targetIdx = secrets.IndexOf(targetKeyword.ToLower(), StringComparison.Ordinal);
             if (secretIdx == -1 && targetTokenIdx == -1 && targetIdx == -1)
             {
                 throw new AppCenterException(parseErrorMessage);
--- a/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
@@ -36,10 +36,10 @@ namespace Microsoft.AppCenter
                 return secrets;
             }
 
-            Dictionary<string, string> secretsDictionary = secrets.Split(SecretDelimiter)
-            .Select(value => value.Split(PlatformKeyValueDelimiter))
-            .GroupBy(p => p[0], StringComparer.OrdinalIgnoreCase)
-            .ToDictionary(pair => pair.Key, pair => pair.Last()[1], StringComparer.OrdinalIgnoreCase);
+            var secretsDictionary = secrets.Split(SecretDelimiter)
+                                           .Select(value => value.Split(PlatformKeyValueDelimiter))
+                                           .GroupBy(p => p[0] ?? "", StringComparer.OrdinalIgnoreCase)
+                                           .ToDictionary(pair => pair.Key, pair => pair.Last().Last(), StringComparer.OrdinalIgnoreCase);
 
             var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
             if (secretsDictionary.ContainsKey(TargetPostfix.ToLower()) || secretsDictionary.ContainsKey(SecretPostfix.ToLower()))
@@ -60,8 +60,10 @@ namespace Microsoft.AppCenter
             {
                 foundTarget = secretsDictionary.TryGetValue(platformTargetIdentifier, out platformTargetToken);
             }
+            var foundSomething = foundTarget || foundSecret;
+            var someKeyExists = !string.IsNullOrEmpty(platformSecret) || !string.IsNullOrEmpty(platformTargetToken);
 
-            if ((!foundSecret && !foundTarget) || (string.IsNullOrEmpty(platformSecret) && string.IsNullOrEmpty(platformTargetToken)))
+            if (!foundSomething || !someKeyExists)
             {
                 throw new AppCenterException(parseErrorMessage);
             }
--- a/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
@@ -11,14 +11,17 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        const char delimiter = ';';
-        const string targetKeyword = "Target";
+        const char Delimiter = ';';
+        const string PlatformIndicator = '=';
+        const string TargetPostfix = "Target";
+        const string SecretPostfix = "appSecret";
 
         // Gets the first instance of an app sceret and/or target token corresponding to the given platform name, or returns the string 
         // as-is if no identifier can be found. Logs a message if no identifiers can be found.
         internal static string GetSecretAndTargetForPlatform(string secrets, string platformIdentifier)
         {
-            var platformTargetIdentifier = platformIdentifier + targetKeyword;
+            var platformTargetIdentifier = platformIdentifier + TargetPostfix
+    ;
             if (string.IsNullOrEmpty(secrets))
             {
                 throw new AppCenterException("App secrets string is null or empty");
@@ -26,7 +29,7 @@ namespace Microsoft.AppCenter
 
             // If there are no equals signs, then there are no named identifiers, but log a message in case the developer made 
             // a typing error.
-            if (!secrets.Contains("="))
+            if (!secrets.Contains(PlatformIndicator))
             {
                 AppCenterLog.Debug(AppCenterLog.LogTag, "No named identifier found in appSecret; using as-is");
                 return secrets;
@@ -34,11 +37,11 @@ namespace Microsoft.AppCenter
 
             var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
 
-            var platformIndicator = platformIdentifier + "=";
-            var platformTargetIdicator = platformTargetIdentifier + "=";
+            var platformIndicator = platformIdentifier + PlatformIndicator;
+            var platformTargetIdicator = platformTargetIdentifier + PlatformIndicator;
             var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
             var targetTokenIdx = secrets.IndexOf(platformTargetIdicator, StringComparison.Ordinal);
-            var targetIdx = secrets.IndexOf(targetKeyword.ToLower(), StringComparison.Ordinal);
+            var targetIdx = secrets.IndexOf(TargetPostfix.ToLower(), StringComparison.Ordinal);
             if (secretIdx == -1 && targetTokenIdx == -1 && targetIdx == -1)
             {
                 throw new AppCenterException(parseErrorMessage);
@@ -58,7 +61,7 @@ namespace Microsoft.AppCenter
             }
             var platformSecret = FindTheKey(secretIdx, secrets);
             var platformTargetToken = FindTheKey(targetTokenIdx, secrets);
-            if (platformSecret == string.Empty && platformTargetToken == string.Empty)
+            if (string.IsNullOrEmpty(platformSecret) && string.IsNullOrEmpty(platformTargetToken))
             {
                 throw new AppCenterException(parseErrorMessage);
             }
@@ -69,9 +72,9 @@ namespace Microsoft.AppCenter
                 //If there is an app secret
                 if (platformSecret.Length > 0)
                 {
-                    platformSecret = "appSecret=" + platformSecret + ";";
+                    platformSecret = SecretPostfix + PlatformIndicator + platformSecret + Delimiter;
                 }
-                platformSecret += "target=" + platformTargetToken;
+                platformSecret += TargetPostfix.toLower() + PlatformIndicator + platformTargetToken;
             }
             return platformSecret;
         }
@@ -84,7 +87,7 @@ namespace Microsoft.AppCenter
                 while (keyIdx < secrets.Length)
                 {
                     var nextChar = secrets[keyIdx++];
-                    if (nextChar == delimiter)
+                    if (nextChar == Delimiter)
                     {
                         break;
                     }
--- a/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
@@ -15,14 +15,14 @@ namespace Microsoft.AppCenter
     {
         const string SecretDelimiter = ";";
         const string PlatformKeyValueDelimiter = "=";
-        const string TargetPostfix = "target";
-        const string SecretPostfix = "appsecret";
+        const string TargetKeyName = "target";
+        const string AppSecretKeyName = "appsecret";
 
         // Gets the first instance of an app sceret and/or target token corresponding to the given platform name, or returns the string 
         // as-is if no identifier can be found. Logs a message if no identifiers can be found.
         internal static string GetSecretAndTargetForPlatform(string secrets, string platformIdentifier)
         {
-            var platformTargetIdentifier = platformIdentifier + TargetPostfix;
+            var platformTargetIdentifier = platformIdentifier + TargetKeyName;
             if (string.IsNullOrEmpty(secrets))
             {
                 throw new AppCenterException("App secrets string is null or empty");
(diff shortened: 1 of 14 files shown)


SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -31,31 +33,70 @@ internal static string GetSecretForPlatform(string secrets, string platformIdent
...
+                    var nextChar = secrets[targetTokenIdx++];
+                    if (nextChar == ';')
+                    {
+                        break;
+                    }
 
-                platformSecret += nextChar;
+                    platformTargetToken += nextChar;
+                }
             }
```
MatkovIvan: Looks like code dup. Move it to a new function

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -31,31 +33,70 @@ internal static string GetSecretForPlatform(string secrets, string platformIdent
             var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
 
             var platformIndicator = platformIdentifier + "=";
+            var platformTargetIdicator = platformTargetIdentifier + "=";
             var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
-            if (secretIdx == -1)
+            var targetTokenIdx = secrets.IndexOf(platformTargetIdicator, StringComparison.Ordinal);
+            if (secretIdx == -1 && targetTokenIdx == -1)
             {
                 throw new AppCenterException(parseErrorMessage);
```
MatkovIvan: I guess the input like `"target={}"` should return as-is without any exception

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,10 +11,14 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const char delimiter = ';';
+        const string targetKeyword = "Target";
```
MatkovIvan:
```suggestion

        const string TargetPostfix = "Target";
```

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,10 +11,14 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const char delimiter = ';';
```
MatkovIvan:
```suggestion
        const char Delimiter = ';';
```

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -31,31 +35,63 @@ internal static string GetSecretForPlatform(string secrets, string platformIdent
             var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
 
             var platformIndicator = platformIdentifier + "=";
+            var platformTargetIdicator = platformTargetIdentifier + "=";
```
MatkovIvan: Move this separator to consts too?

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -31,31 +35,63 @@ internal static string GetSecretForPlatform(string secrets, string platformIdent
...
             var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
-            if (secretIdx == -1)
+            var targetTokenIdx = secrets.IndexOf(platformTargetIdicator, StringComparison.Ordinal);
+            var targetIdx = secrets.IndexOf(targetKeyword.ToLower(), StringComparison.Ordinal);
+            if (secretIdx == -1 && targetTokenIdx == -1 && targetIdx == -1)
+            {
+                throw new AppCenterException(parseErrorMessage);
+            }
+            if (targetIdx >= 0 && secretIdx == -1 && targetTokenIdx == -1)
```
MatkovIvan: why not `!= -1`?

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -31,31 +35,63 @@ internal static string GetSecretForPlatform(string secrets, string platformIdent
...
+            {
+                secretIdx += platformIndicator.Length;
+            }
+            if (targetTokenIdx >= 0)
+            {
+                targetTokenIdx += platformTargetIdicator.Length;
+            }
+            var platformSecret = FindTheKey(secretIdx, secrets);
+            var platformTargetToken = FindTheKey(targetTokenIdx, secrets);
+            if (platformSecret == string.Empty && platformTargetToken == string.Empty)
```
MatkovIvan: https://docs.microsoft.com/en-us/dotnet/api/system.string.isnullorempty

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -31,31 +35,63 @@ internal static string GetSecretForPlatform(string secrets, string platformIdent
...
-                if (nextChar == ';')
+                //If there is an app secret
+                if (platformSecret.Length > 0)
                 {
-                    break;
+                    platformSecret = "appSecret=" + platformSecret + ";";
                 }
-
-                platformSecret += nextChar;
+                platformSecret += "target=" + platformTargetToken;
```
MatkovIvan: const?

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
annakocheshkova: Because we let that pass only if there's a string containing 'target='

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,34 +11,37 @@ namespace Microsoft.AppCenter
...
+        const string TargetPostfix = "Target";
+        const string SecretPostfix = "appSecret";
 
         // Gets the first instance of an app sceret and/or target token corresponding to the given platform name, or returns the string 
         // as-is if no identifier can be found. Logs a message if no identifiers can be found.
         internal static string GetSecretAndTargetForPlatform(string secrets, string platformIdentifier)
         {
-            var platformTargetIdentifier = platformIdentifier + targetKeyword;
+            var platformTargetIdentifier = platformIdentifier + TargetPostfix
+    ;
```
MatkovIvan: Format it

annakocheshkova: Oh

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,51 +11,89 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const char Delimiter = ';';
+        const string PlatformIndicator = "=";
+        const string TargetPostfix = "Target";
```
guperrot: Postfix is not the right term, use Suffix.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,51 +11,89 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const char Delimiter = ';';
+        const string PlatformIndicator = "=";
```
guperrot: PlatformKeyValueDelimiter

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,51 +11,89 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const char Delimiter = ';';
```
guperrot: SecretDelimiter

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,51 +11,89 @@ namespace Microsoft.AppCenter
...
+            if (secretIdx == -1 && targetTokenIdx == -1 && targetIdx == -1)
+            {
+                throw new AppCenterException(parseErrorMessage);
+            }
+            if (targetIdx >= 0 && secretIdx == -1 && targetTokenIdx == -1) 
+            {
+                AppCenterLog.Debug(AppCenterLog.LogTag, "Found named identifier 'target' in the secret. Returning as-is.");
+                return secrets;
+            }
+            if (secretIdx >= 0)
```
guperrot: This code is too complex for what it's doing and hard to read.
Let's use Split and Linq, example: https://stackoverflow.com/a/11673966

guperrot: To combine with https://stackoverflow.com/a/3319086 in case there are duplicate keys.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,71 @@ namespace Microsoft.AppCenter
...
+            if (!secrets.Contains(PlatformKeyValueDelimiter.ToString()))
             {
                 AppCenterLog.Debug(AppCenterLog.LogTag, "No named identifier found in appSecret; using as-is");
                 return secrets;
             }
 
-            var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
+            var secretsDictionary = secrets.Split(SecretDelimiter)
+                                           .Select(value => value.Split(PlatformKeyValueDelimiter))
+                                           .GroupBy(p => p[0].Trim() ?? "", StringComparer.OrdinalIgnoreCase)
```
dhei: The `GroupBy()` and `ToDictionary()` LINQ query is a little bit challenging for me to understand, perticulaly the `p => p[0].Trim()` part and `pair => pair.Key.Trim(), pair => pair.Last().Last().Trim()` part.

It would be nice if we could break this LINQ statement into something mutiple-step that is easier to follow.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,71 @@ namespace Microsoft.AppCenter
...
+                                           .ToDictionary(pair => pair.Key.Trim(), pair => pair.Last().Last().Trim(), StringComparer.OrdinalIgnoreCase);
 
-            var platformIndicator = platformIdentifier + "=";
-            var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
-            if (secretIdx == -1)
+            var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
+            if (secretsDictionary.ContainsKey(TargetPostfix.ToLower()) || secretsDictionary.ContainsKey(SecretPostfix.ToLower()))
             {
-                throw new AppCenterException(parseErrorMessage);
+                AppCenterLog.Debug(AppCenterLog.LogTag, "Found named identifier in the secret. Returning as-is.");
```
dhei: Align the message with https://github.com/microsoft/appcenter-sdk-dotnet/pull/1176/files#diff-390008ed19e9532de33a029612f1221fR35.
           
```suggestion
                AppCenterLog.Debug(AppCenterLog.LogTag, "Found named identifier in the secret; using as-is.");
```

Tests/Microsoft.AppCenter.Test.Windows/AppCenterTest.cs
```diff
@@ -515,10 +515,68 @@ public void ParseAppSecretOnePlatformTerminatingSemicolon()
...
-            var parsedSecret = AppCenter.GetSecretForPlatform(secrets, platformId);
+            var parsedSecret = AppCenter.GetSecretAndTargetForPlatform(secrets, platformId);
             Assert.AreEqual(appSecret, parsedSecret);
         }
 
+        /// <summary>
+        /// Verify parse when there is only one platform and both app secret and token
+        /// </summary>
+        [TestMethod]
+        public void ParseAppSecretAndTargetOnePlatformTerminatingSemicolon()
```
dhei: The secret string doesn't contain a terminating semicolon.
```suggestion
        public void ParseAppSecretAndTargetOnePlatform()
```

Tests/Microsoft.AppCenter.Test.Windows/AppCenterTest.cs
```diff
@@ -515,10 +515,68 @@ public void ParseAppSecretOnePlatformTerminatingSemicolon()
...
+            var parsedSecret = AppCenter.GetSecretAndTargetForPlatform(secrets, platformId);
+            var expected = $"appSecret={appSecret};target={targetToken}";
+            Assert.AreEqual(expected, parsedSecret);
+        }
+
+        /// <summary>
+        /// Verify parse when there is several platforms and both app secret and token
+        /// </summary>
+        [TestMethod]
+        public void ParseAppSecretAndTargetMultiplePlatformTerminatingSemicolon()
```
dhei: The secret string doesn't contain a terminating semicolon.

```suggestion
        public void ParseAppSecretAndTargetMultiplePlatform()
```

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -28,17 +28,17 @@ internal static string GetSecretAndTargetForPlatform(string secrets, string plat
...
                 return secrets;
             }
 
             var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
 
-            var platformIndicator = platformIdentifier + PlatformIndicator;
-            var platformTargetIdicator = platformTargetIdentifier + PlatformIndicator;
-            var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
+            var PlatformKeyValueDelimiter = platformIdentifier + PlatformKeyValueDelimiter;
+            var platformTargetIdicator = platformTargetIdentifier + PlatformKeyValueDelimiter;
```
michaelsmitha: `platformTargetIdicator` - `Indicator` has a typo

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,71 @@ namespace Microsoft.AppCenter
...
+            }
+            if (secretsDictionary.ContainsKey(platformTargetIdentifier))
+            {
+                foundTarget = secretsDictionary.TryGetValue(platformTargetIdentifier, out platformTargetToken);
             }
+            var foundNoneOfTheKeys = !foundTarget && !foundSecret;
+            var bothKeysAreEmpty = string.IsNullOrEmpty(platformSecret) && string.IsNullOrEmpty(platformTargetToken);
 
-            if (platformSecret == string.Empty)
+            if (foundNoneOfTheKeys || bothKeysAreEmpty)
```
dhei: I don't see a unit test for bothKeysAreEmpty condition. Can you add one?

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,71 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const char SecretDelimiter = ';';
```
dhei: These 4 const can be `private`, right?

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,71 @@ namespace Microsoft.AppCenter
...
+            var platformTargetIdentifier = platformIdentifier + TargetPostfix;
             if (string.IsNullOrEmpty(secrets))
             {
                 throw new AppCenterException("App secrets string is null or empty");
             }
 
             // If there are no equals signs, then there are no named identifiers, but log a message in case the developer made 
             // a typing error.
-            if (!secrets.Contains("="))
+            if (!secrets.Contains(PlatformKeyValueDelimiter.ToString()))
```
dhei: `PlatformKeyValueDelimiter` can be declared as string instead of char because C# string split accept string as well. Then you don't have to `.ToString()`.

dhei: Same goes to char `SecretDelimiter`.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
guperrot: I think the default visibility is private here, and if you add private then Xamarin Studio will suggest you to remove it.

dhei: image
I don't think the default access modifier of C# const is `private`.  And you can see that VS for Mac does allow me to add `private`  here.

guperrot: https://stackoverflow.com/a/3763638/598520

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,74 @@ namespace Microsoft.AppCenter
...
+                                           .Select(value => value.Split(PlatformKeyValueDelimiter.ToCharArray()))
+                                           .GroupBy(p => p[0].Trim() ?? "", StringComparer.OrdinalIgnoreCase);
+
+            // Create a dictionary choosing the last secret value for each key.
+            var secretsDictionary = secretsGroup.ToDictionary(pair => pair.Key.Trim(), pair => pair.Last().Last().Trim(), StringComparer.OrdinalIgnoreCase);
 
-            var platformIndicator = platformIdentifier + "=";
-            var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
-            if (secretIdx == -1)
+            var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
```
michaelsmitha: `parseErrorMessage` is only used on line 71 now, doesn't need its own variable.

Tests/Microsoft.AppCenter.Test.Windows/AppCenterTest.cs
```diff
@@ -515,10 +515,97 @@ public void ParseAppSecretOnePlatformTerminatingSemicolon()
...
+            var targetToken = Guid.NewGuid().ToString();
+            var platformId = "android";
+            var secrets = $"{platformId}Target={targetToken};";
+            var parsedSecret = AppCenter.GetSecretAndTargetForPlatform(secrets, platformId);
+            var expected = $"target={targetToken}";
+            Assert.AreEqual(expected, parsedSecret);
+        }
+
+        /// <summary>
+        /// Verify parse when there is only token
```
michaelsmitha: The summary of this test is the same as the one above, doesn't describe the test

Tests/Microsoft.AppCenter.Test.Windows/AppCenterTest.cs
```diff
@@ -515,10 +515,97 @@ public void ParseAppSecretOnePlatformTerminatingSemicolon()
...
+            var parsedSecret = AppCenter.GetSecretAndTargetForPlatform(secrets, platformId);
+            var expected = $"appSecret={appSecret};target={targetToken}";
+            Assert.AreEqual(expected, parsedSecret);
+        }
+
+        /// <summary>
+        /// Verify parse when there is only token
+        /// </summary>
+        [TestMethod]
+        public void ParseTargetTerminatingSemicolon()
```
michaelsmitha: The summary is about testing only the token, the test name is about terminating semicolon?

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,74 @@ namespace Microsoft.AppCenter
...
+                return secrets;
             }
-            secretIdx += platformIndicator.Length;
-            var platformSecret = string.Empty;
 
-            while (secretIdx < secrets.Length)
+            var platformSecret = string.Empty;
+            var platformTargetToken = string.Empty;
+            bool foundSecret = false;
+            bool foundTarget = false;
```
achocron: please use `var`

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,74 @@ namespace Microsoft.AppCenter
...
+                                           .GroupBy(p => p[0].Trim() ?? "", StringComparer.OrdinalIgnoreCase);
+
+            // Create a dictionary choosing the last secret value for each key.
+            var secretsDictionary = secretsGroup.ToDictionary(pair => pair.Key.Trim(), pair => pair.Last().Last().Trim(), StringComparer.OrdinalIgnoreCase);
 
-            var platformIndicator = platformIdentifier + "=";
-            var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
-            if (secretIdx == -1)
+            var parseErrorMessage = $"Error parsing key for '{platformIdentifier}'";
+            if (secretsDictionary.ContainsKey(TargetPostfix.ToLower()) || secretsDictionary.ContainsKey(SecretPostfix.ToLower()))
```
guperrot: Please check native SDKs, if we are case sensitive there, we should do the same here.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,74 @@ namespace Microsoft.AppCenter
...
-                }
-
-                platformSecret += nextChar;
+                foundSecret = secretsDictionary.TryGetValue(platformIdentifier, out platformSecret);
+            }
+            if (secretsDictionary.ContainsKey(platformTargetIdentifier))
+            {
+                foundTarget = secretsDictionary.TryGetValue(platformTargetIdentifier, out platformTargetToken);
             }
+            var foundNoneOfTheKeys = !foundTarget && !foundSecret;
```
guperrot: we don't need booleans since we have nullable strings already.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
guperrot: Also I am reading that wrong are the exclamation marks missing?

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,74 @@ namespace Microsoft.AppCenter
...
+            var bothKeysAreEmpty = string.IsNullOrEmpty(platformSecret) && string.IsNullOrEmpty(platformTargetToken);
 
-            if (platformSecret == string.Empty)
+            if (foundNoneOfTheKeys || bothKeysAreEmpty)
             {
                 throw new AppCenterException(parseErrorMessage);
             }
 
+            // Format the string as "appSecret={};target={}" or "target={}" if needed.
+            if (platformTargetToken.Length > 0)
```
guperrot: Keep using string.IsNullOrEmpty

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
annakocheshkova: Do you mean invert the condition? It's correct. We should return "target=123" or "appSecret=123" as is.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,66 @@ namespace Microsoft.AppCenter
...
+            // Grouping by the name of the key.
+            var secretsGroup = secrets.Split(SecretDelimiter.ToCharArray())
+                                           .Select(value => value.Split(PlatformKeyValueDelimiter.ToCharArray()))
+                                           .GroupBy(p => p[0].Trim() ?? "", StringComparer.OrdinalIgnoreCase);
 
-            var platformIndicator = platformIdentifier + "=";
-            var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
-            if (secretIdx == -1)
+            // Create a dictionary choosing the last secret value for each key.
+            var secretsDictionary = secretsGroup.ToDictionary(pair => pair.Key.Trim(), pair => pair.Last().Last().Trim(), StringComparer.OrdinalIgnoreCase);
```
jwallra: `pair.Last().Last().Trim()` is a bit weird because of the two `Last()` calls, can you explain?  Maybe add a comment

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,66 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const string SecretDelimiter = ";";
+        const string PlatformKeyValueDelimiter = "=";
+        const string TargetPostfix = "target";
+        const string SecretPostfix = "appsecret";
```
jwallra: These are prefixes, not postfixes

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,66 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const string SecretDelimiter = ";";
+        const string PlatformKeyValueDelimiter = "=";
+        const string TargetPostfix = "target";
```
dhei: Given the format of `"appSecret=123;target=123"`, I don't think they are "postfix" at all. 

Maybe something like
```
```

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,66 @@ namespace Microsoft.AppCenter
...
 
-            var platformIndicator = platformIdentifier + "=";
-            var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
-            if (secretIdx == -1)
+            // Create a dictionary choosing the last secret value for each key. If the key has more than one secret value then select the last of them.
+            var secretsDictionary = secretsGroup.ToDictionary(pair => pair.Key.Trim(), pair => pair.Last().Last().Trim(), StringComparer.OrdinalIgnoreCase);
+            if (secretsDictionary.ContainsKey(TargetKeyName) || secretsDictionary.ContainsKey(AppSecretKeyName))
             {
-                throw new AppCenterException(parseErrorMessage);
+                AppCenterLog.Debug(AppCenterLog.LogTag, "Found named identifier in the secret; using as-is.");
```
guperrot: I still don't understand this.
If we find named identifiers, we use as is, I don't understand the message being logged.
This needs to be clarified.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,66 @@ namespace Microsoft.AppCenter
...
+            // Grouping by the name of the key.
+            var secretsGroup = secrets.Split(SecretDelimiter.ToCharArray())
+                                           .Select(value => value.Split(PlatformKeyValueDelimiter.ToCharArray()))
+                                           .GroupBy(p => p[0].Trim() ?? "", StringComparer.OrdinalIgnoreCase);
 
-            var platformIndicator = platformIdentifier + "=";
-            var secretIdx = secrets.IndexOf(platformIndicator, StringComparison.Ordinal);
-            if (secretIdx == -1)
+            // Create a dictionary choosing the last secret value for each key. If the key has more than one secret value then select the last of them.
+            var secretsDictionary = secretsGroup.ToDictionary(pair => pair.Key.Trim(), pair => pair.Last().Last().Trim(), StringComparer.OrdinalIgnoreCase);
```
guperrot: I was sure I commented on this but maybe not.
We are case sensitive in Android so please make it consistent with the native SDK, not ignoring case.

guperrot:
Why all the trimming? Our native SDKs don't do that.

annakocheshkova:
@guperrot some tests have spaces in strings. Without trimming they fail.

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -11,50 +13,72 @@ namespace Microsoft.AppCenter
     /// </summary>
     public partial class AppCenter
     {
-        // Gets the first instance of an app secret corresponding to the given platform name, or returns the string 
+        const string SecretDelimiter = ";";
```
guperrot: Was this tested on all windows platforms as well to make sure since this is shared code?

guperrot: And both UWP apps (the forms and non forms).

AnastasiaKubova: yes, it works expected on windows and uwp platforms

PR #1208, opened 2019-11-11
Configuration actualization before settings value changing

olmisa:

## Description

Configuration reopened before settings new value saving.

## Misc

[AB#72324](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/72324)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@
 
 * **[Feature]** Support sending attachments in handled errors.
 
+* **[Fix]** Fix an error on running same application twice and enabling App Center.
+
 ### App Center Distribute
 
 #### Android
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # App Center SDK for .NET Change Log
 
+## Version 2.6.2
+
+### App Center
+
+#### UWP/WPF/WinForms
+
+* **[Fix]** Fix an error on running same application twice and enabling App Center.
+
+___
+
 ## Version 2.6.1
 
 ### App Center
@@ -21,8 +31,6 @@
 #### UWP/WPF/WinForms
 
 * **[Feature]** Support sending attachments in handled errors.
-
-* **[Fix]** Fix an error on running same application twice and enabling App Center.
 
 ### App Center Distribute
 
--- a/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DefaultApplicationSettings.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DefaultApplicationSettings.cs
@@ -59,6 +59,7 @@ namespace Microsoft.AppCenter.Utils
         {
             lock (configLock)
             {
+                configuration = OpenConfiguration();
                 configuration.AppSettings.Settings.Remove(key);
                 configuration.Save();
             }
--- a/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DefaultApplicationSettings.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DefaultApplicationSettings.cs
@@ -68,6 +68,7 @@ namespace Microsoft.AppCenter.Utils
         {
             lock (configLock)
             {
+                configuration = OpenConfiguration();
                 var element = configuration.AppSettings.Settings[key];
                 if (element == null)
                 {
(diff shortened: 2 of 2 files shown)


SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DefaultApplicationSettings.cs
```diff
@@ -68,6 +68,7 @@ private void SaveValue(string key, string value)
         {
             lock (configLock)
             {
+                configuration = OpenConfiguration();
```
guperrot: Wait are we trying to open the file for every single property change?
Do we have issues with read?

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 
 * **[Feature]** Support sending attachments in handled errors.
 
+* **[Fix]** Fix an error on running same application twice and enabling App Center.
```
guperrot: This needs to be a new version 2.6.2

guperrot: Please catch up with develop: https://github.com/microsoft/appcenter-sdk-dotnet/pull/1210/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1e

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DefaultApplicationSettings.cs
jwallra: Also, do we have to release the file in some way?  I would expect that this would prevent writing saving values in the second application at all

guperrot: We will parking lot this as this can cause a lot of side effects with inconsistent states between saved. Like A disable only crashes, B disables only Push, what does B save?

olmisa: We have several (same) applications that make changes in the same file. As I understand, the config file was opened for writing only when the 'Configuration.Save' method works.
So we have the following situation. The first app changed the config file. In this case, the second app can't save changes without configuration refreshing
(we see the exception and that is why I have added refreshing current configuration from file). In this case, the configuration should be refreshed each time when we save changes (before 'Save' method called). 
Because when the first app made changes in the file then the second app should refresh configuration. After that, the second app saves other changes to the file and the first app configuration should be refreshed or the same exception will appear in the first app.
If the first app disables only 'Crashes' and after that, the second app disables only 'Push' then the 'Crashes' and the 'Push' both will be disabled in the second application.
Maybe we should decide what solution is better:
1. All app instances could write the app config but in this case, this is a problem to check that all app instances have the right configuration (maybe this is a problem of the application authors).
2. Only one application instance could write to our configuration file and when the second app tries to write some changes then we write some logs and ignore these changes.
Maybe there are other ways to solve this problem.

CHANGELOG.md
olmisa: Fixed

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DefaultApplicationSettings.cs
guperrot: We decided to not reopen the file every time.
See our internal conversation thread for details.

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DefaultApplicationSettings.cs
```diff
@@ -77,8 +77,20 @@ private void SaveValue(string key, string value)
...
+
+        private void SaveConfiguration()
+        {
+            try
+            {
                 configuration.Save();
             }
+            catch (ConfigurationErrorsException e)
+            {
+                AppCenterLog.Warn(AppCenterLog.LogTag, $"Configuration file can't be saved. Failure reason: {e.Message}");
```
guperrot: Looks good but make sure to test SQLite interactions from different processes before we merge.
We might need to spam more logs to see issues (try a for loop).

olmisa: Done. I have created 'for' loop from 0 to 1000000. In loop 'Analytics.TrackEvent' method was called. It was done with two active app instances. No exceptions.

olmisa: Update, SQLite Busy error was reproduced. 28000-30000 logs were tracked as usual, after that the error could appear.

guperrot: What happens after the error? Is the SDK printing it? Is the app crashing? Can you track more events correctly after the errors or is not recoverable?

Issue #1184, opened 2019-10-30
SQLite exception sending crashes in UWP


asierpn:
### **Description**

I've enabled the Crashes SDK in our UWP project, our tests in debug and release with sideloading were succesfull, but after publishing a new version in Windows Store we have started to see this exception in HockeyApp:

> Date/Time:       2019-10-29T19:59:23.1704098Z
> OS Version:      Windows 10.0.18362.418
> Report Version:  104

> Exception Type:  System.AggregateException
> Crashed Thread:  13

> Application Specific Information:
> A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Corrupt)

Here is the stack trace of the exception
[stacktrace.txt](https://github.com/microsoft/appcenter-sdk-dotnet/files/3787247/stacktrace.txt)

The TrackEvent API of the Analytics SDK is working fine.

### **Repro Steps**

1. Enable App Center Crashes and Analytics in UWP project
2. Call Crashes.TrackError passing a handled exception and some additional properties
3. Publish in Microsoft Store and execute this version

annakocheshkova:
Hi there, thank you for the report! Do you have any information about the specific types of devices this issue happens on or any environment info?

asierpn:
We are publishing a new version in Microsoft Store with an option to manually force an exception to try to reproduce the issue.

At this moment this has happened only to one user yesterday using an HP ProDesk 400 G3 from Windows 10.0.18362.418, the SDK version is 1.5.0. The UWP project targets 18362 with minimum version 17763. The AppCenter nuget package is installed in a UWP class library project.

There is 58 crash logs of the same issue and same user with 5 minutes of difference between them.

annakocheshkova:
Am I correct is this 1.5.0 (not 2.5.0)? Any particular reason you are not updating to the newer versions?

asierpn:
Oops, sorry for the mistake, it's version 2.5.0

asierpn:
We have tried to reproduce the issue on other computers and it seems that the issue is only happening on a concrete computer, I've physical access to this computer as it is an internal computer, is there any check we can do on its file system to determine the cause of this exception?

guperrot:
Do you have a dependency on SQLite? If yes what package are you using with what version? It might be a conflict with the one we depend on from the SDK.

guperrot:
[...]

PR #1214, opened 2019-11-14
Delete corrupted database file

guperrot:

## Description

Delete corrupted databases and create a new one.

## Related PRs or issues

#1184 part 2
[AB#71860](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/71860)

## Misc

The DB deletion does not work yet. To investigate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 #### UWP/WPF/WinForms
 
 * **[Fix]** Fix an issue where internal SDK exceptions would sometimes be "unobserved" and would cause undefined behavior.
+* **[Fix]** Fix deleting the database file and opening a new one whenever corruption is detected.
+
 ___
 
 ## Version 2.6.1
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 #### UWP/WPF/WinForms
 
 * **[Fix]** Fix an issue where internal SDK exceptions would sometimes be "unobserved" and would cause undefined behavior.
-* **[Fix]** Fix deleting the database file and opening a new one whenever corruption is detected.
+* **[Fix]** The SDK now detects the SQLite database is corrupted and deletes it and creates a new one to start over.
 
 ___
 
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
@@ -383,11 +383,11 @@ namespace Microsoft.AppCenter.Storage
             // that it's not always originated by a proper SQLiteException (which would then be converted to StorageException in StorageAdapter).
             // If it was always the right type then the exception would not have been unobserved in that application before we changed the re-throw logic here.
             // But the message is definitely "Corrupt" and thus unfortunately that is the only check we seem to be able to do as opposed to type/property checking.
-            if (e.Message == "Corrupt")
+            if (e.Message == "Corrupt" || e.InnerException?.Message == "Corrupt")
             {
                 AppCenterLog.Error(AppCenterLog.LogTag, "Database corruption detected, deleting the file and starting fresh...", e);
                 await _storageAdapter.DeleteDatabaseFileAsync();
-                //await InitializeDatabaseAsync();
+                await InitializeDatabaseAsync();
             }
 
             // Return exception to re-throw.
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
@@ -391,14 +391,16 @@ namespace Microsoft.AppCenter.Storage
             }
 
             // Return exception to re-throw.
-            if (!(e is StorageException))
+            if (e is StorageException)
             {
-                // Tasks should already be throwing only storage exceptions, but in case any are missed, 
-                // which has happened (the Corrupt exception mentioned previously), catch them here and wrap in a storage exception. This will prevent 
-                // the exception from being unobserved.
-                return new StorageException(e);
+                // This is the expected case, storage adapter already wraps exception as StorageException, so return as is.
+                return e;
             }
-            return e;
+
+            // Tasks should already be throwing only storage exceptions, but in case any are missed, 
+            // which has happened (the Corrupt exception mentioned previously), catch them here and wrap in a storage exception. This will prevent 
+            // the exception from being unobserved.
+            return new StorageException(e);
         }
 
         private void AddTaskToQueue(Task task)
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
@@ -386,8 +386,8 @@ namespace Microsoft.AppCenter.Storage
             if (e.Message == "Corrupt" || e.InnerException?.Message == "Corrupt")
             {
                 AppCenterLog.Error(AppCenterLog.LogTag, "Database corruption detected, deleting the file and starting fresh...", e);
-                await _storageAdapter.DeleteDatabaseFileAsync();
-                await InitializeDatabaseAsync();
+                await _storageAdapter.DeleteDatabaseFileAsync().ConfigureAwait(false);
+                await InitializeDatabaseAsync().ConfigureAwait(false);
             }
 
             // Return exception to re-throw.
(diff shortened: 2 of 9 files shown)


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -349,12 +349,10 @@ private Task AddTaskToQueue(Action action)
...
                     action();
                 }
-                catch (Exception e) when (!(e is StorageException))
+                catch (Exception e)
                 {
-                    // Tasks should already be throwing only storage exceptions, but in case any are missed, 
-                    // which has happened, catch them here and wrap in a storage exception. This will prevent 
-                    // the exception from being unobserved.
-                    throw new StorageException(e);
+                    // If we use await on a task that returns an exception, it makes the app crash, however, using the awaiter works...
```
MatkovIvan: Is it documented behavior?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -369,18 +367,40 @@ private Task<T> AddTaskToQueue<T>(Func<T> action)
...
             return task;
         }
 
+        private async Task<Exception> HandleStorageRelatedExceptionAsync(Exception e)
+        {
+            // Check if database is corrupted, we have evidence (https://github.com/microsoft/appcenter-sdk-dotnet/issues/1184)
+            // that it's not always originated by a proper SQLiteException (which would then be converted to StorageException in StorageAdapter).
+            // If it was always the right type then the exception would not have been unobserved in that application before we changed the re-throw logic here.
+            // But the message is definitely "Corrupt" and thus unfortunately that is the only check we seem to be able to do as opposed to type/property checking.
+            if (e.Message == "Corrupt")
```
MatkovIvan: Do we sure that case-sensitive comparison is enough? Maybe we should use "contains"?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -369,18 +367,40 @@ private Task<T> AddTaskToQueue<T>(Func<T> action)
...
+            {
+                AppCenterLog.Error(AppCenterLog.LogTag, "Database corruption detected, deleting the file and starting fresh...", e);
+                await _storageAdapter.DeleteDatabaseFileAsync();
+                await InitializeDatabaseAsync();
+            }
+
+            // Return exception to re-throw.
+            if (!(e is StorageException))
```
MatkovIvan: Tried to find how to simplify it (drop `()`) and found [this](https://stackoverflow.com/questions/529944/check-if-object-is-not-of-type-equivalent-for-is-c-sharp#comment67093273_529959)
Interesting variant, not sure what's better. 
BTW Kotlin has [a nice construction for this case](https://kotlinlang.org/docs/reference/typecasts.html#is-and-is-operators) - `!is`

```suggestion
            if (false == e is StorageException)
```

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/Files/File.cs
```diff
@@ -20,6 +20,11 @@ public File()
         {
         }
 
+        public File(string filePath)
```
MatkovIvan:
```suggestion
        public File(string filePath) : this(new FileInfo(filePath))
```

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/Files/File.cs
```diff
@@ -20,6 +20,11 @@ public File()
...
+        {
+            _underlyingFileInfo = new FileInfo(filePath);
```
MatkovIvan:
```suggestion
```

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
guperrot: I don't see this as an improvement.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
guperrot: Right now all occurrences I found while searching for this is this specific casing.
Now the check needs to be improved as it could be a legit inner SQLiteException, so need to check inner on 1 level as well.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
guperrot: What I meant by that is, change this line by `throw await HandleStorageRelatedExceptionAsync(e)` and that will crash the app instead of rethrowing.

I don't know why. But anyway the throw await does not compile on the Func variant.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
guperrot: Reversing logic sounds better, will push that instead.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
guperrot: Capture

For the func variant.

guperrot: In the action variant, it crashed instead of being caught:

```
"exception": {
    "type": "Microsoft.AppCenter.Storage.StorageException",
    "message": "The storage operation failed",
    "stackTrace": "   at Microsoft.AppCenter.Storage.Storage.<>c__DisplayClass24_0.<<AddTaskToQueue>b__0>d.MoveNext()\r\n   at Windows.ApplicationModel.Core.UnhandledError.Propagate()\r\n   at Microsoft.AppCenter.Utils.ApplicationLifecycleHelper.<.ctor>b__17_1(Object sender, UnhandledErrorDetectedEventArgs eventArgs)",
    "innerExceptions": [
      {
        "type": "System.Exception",
        "message": "test",
        "stackTrace": "   at Microsoft.AppCenter.Storage.Storage.<>c__DisplayClass24_0.<<AddTaskToQueue>b__0>d.MoveNext()"
      }
    ]
  },
```

guperrot:
a23aac4 tested on uwp/winforms/wpf in app (with fake exceptions).

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
achocron: Did you try `throw (await HandleStorageRelatedExceptionAsync(e))`? Sometimes compound statements with `await` require parentheses

guperrot: I tried, and it offers to remove parentheses, does not change.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,13 +1,14 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using Microsoft.AppCenter.Utils;
```
michaelsmitha: I thought System* was first in sort order

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 #### UWP/WPF/WinForms
 
 * **[Fix]** Fix an issue where internal SDK exceptions would sometimes be "unobserved" and would cause undefined behavior.
+* **[Fix]** Fix deleting the database file and opening a new one whenever corruption is detected.
```
michaelsmitha: the `Fix` above this line states what is being fixed, but this line says `Fix`, then describes what the fix is - but it reads like the deleting and opening a new one is the bug being fixed.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
guperrot: We use the default settings of Visual Studio now to be more friendly to external contributors.

CHANGELOG.md
guperrot: I'll try a different wording.

PR #1236, opened 2019-12-03
Coverage tests for database implementation

AnastasiaKubova:
Things to consider before you submit the PR:

* [ ] Has `CHANGELOG.md` been updated?
* [x] Are tests passing locally?
* [x] Are the files formatted correctly?
* [x] Did you add unit tests if this modifies the Windows code?
* [ ] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

Cover new and changed code with tests.

## Related PRs or issues

[AB#73525](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/73525)

Issue #1134, opened 2019-10-01
App Center WPF running in debug


james1301:
### **Description**

How do I get the App Center WPF running in debug? It is making it frustrating to test when it doesn't seem to work in debug mode.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Start App Center with crashes
2. Display a window on Crashes.ShouldAwaitUserConfirmation
3. Force a crash in your application
4. Restart Application and it won't bring up the crash window

### **Details**

1. What is your app platform (Xamarin.Android or Xamarin.iOS or UWP)?
    - .net framework WPF
2. If using Xamarin.Forms or if using portable/shared code to call our SDK APIs, are you using shared project, PCL code or .NET standard code for the application? Which .NET standard version or which PCL profile?
   N/A
2. Which SDK version are you using?
    - e.g. 2.4.0
4. Which OS version did you experience the issue on?
    - Windows 10 (1903)
5. What device version did you see this error on?  Were you using an emulator or a physical device?
    - Dell PC
6. What third party libraries are you using? 
    - Prism/DevExpress, others...
7. Please enable verbose logging for your app using `AppCenter.LogLevel = LogLevel.Verbose` before your call to `AppCenter.Start(...)` and include the logs here:
2019-10-02 15:33:34.856 [AppCenter] DEBUG: No named identifier found in appSecret; using as-is
2019-10-02 15:33:34.870 [AppCenter] DEBUG: Enabled TLS 1.2 explicitly as it was disabled.
2019-10-02 15:33:34.880 [AppCenter] DEBUG: AddChannel(core)
2019-10-02 15:33:35.082 [AppCenter] INFO: App Center SDK configured successfully.
2019-10-02 15:33:35.087 [AppCenter] DEBUG: AddChannel(crashes)
2019-10-02 15:33:35.091 [AppCenter] INFO: 'Crashes' service started.
2019-10-02 15:33:35.096 [AppCenter] DEBUG: AddChannel(analytics)
2019-10-02 15:33:35.100 [AppCenterAnalytics] DEBUG: SessionTracker.Resume
2019-10-02 15:33:35.106 [AppCenter] DEBUG: Invoke EnqueuingLog event for channel 'analytics'
2019-10-02 15:33:35.459 [AppCenter] DEBUG: Invoke FilteringLog event for channel 'analytics'
2019-10-02 15:33:35.465 [AppCenter] INFO: 'Analytics' service started.
[...]

Issue #1092, opened 2019-08-19
Strong name signing NuGet packages


Suplanus:
In some cases I am only able to use strong name signed DLLs.

I also think it should be standard @ Microsoft:
https://github.com/dotnet/corefx/blob/master/Documentation/project-docs/strong-name-signing.md

EatonZ:
+1 for this. Had to sign them manually as a workaround.

EatonZ:
The [sqlite-net project](https://github.com/praeclarum/sqlite-net) will also need to strong name sign their dll for everything to work (everything silently fails in the background if not signed), [link to issue](https://github.com/praeclarum/sqlite-net/issues/367).

NunoPeixotoPrimavera:
Hi, does anyone have the assemblies and dependencies signed and working?

NunoPeixotoPrimavera:
> The [sqlite-net project](https://github.com/praeclarum/sqlite-net) will also need to strong name sign their dll for everything to work (everything silently fails in the background if not signed), [link to issue](https://github.com/praeclarum/sqlite-net/issues/367).

Hi EatonZ, do you have de assemblies signed and working. Can you send me the dlls?
Or create an attachment on this issue.

EatonZ:
I've attached strong named assemblies for the latest versions. What you want to do is go into your packages folder next to your project's .sln file and replace the dlls in each package's net45 folder (except sqlite-net, replace the one in netstandard2.0).

Any time these NuGet packages are updated or redownloaded/restored, you'll need to do this again.

How I did it: Add [this package](https://www.nuget.org/packages/Brutal.Dev.StrongNameSigner/) to your project and build it. It'll strong name sign any unsigned dlls. Then, replace them in the packages folder, then remove Brutal.Dev.StrongNameSigner from your project.

[Strong Named DLLs.zip](https://github.com/microsoft/appcenter-sdk-dotnet/files/3691485/Strong.Named.DLLs.zip)

NunoPeixotoPrimavera:
Thanks a lot Eaton.

EatonZ:
[sqlite-net-pcl is being removed](https://github.com/microsoft/appcenter-sdk-dotnet/pull/1246) so hopefully this will get resolved next.👍

304NotModified:
All Microsoft dlls are strong named? Why not this one?

jozefizso:
When this will be fixed?

MatkovIvan:
[...]

PR #1246, opened 2019-12-05
Remove sqlite-net-pcl wrapper

MatkovIvan:

## Description

Replace `sqlite-net-pcl` to `SQLitePCLRaw.bundle_green`

## Related PRs or issues

[AB#72743](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/72743)


Tests/Microsoft.AppCenter.Test.Windows/Storage/FakeStorageTest.cs
```diff
@@ -98,6 +101,7 @@ public void GetLogsQueryError()
         /// <summary>
         /// Verify that storage throws StorageException if something went wrong
         /// </summary>
+        /* FIXME
         [TestMethod]
```
MatkovIvan: Test should be fixed in #1236

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
+            }
+            if (result != raw.SQLITE_OK)
+            {
+                var errorMessage = raw.sqlite3_errmsg(_db);
+                throw new StorageException($"Failed to open database connection, result={result}\n\t{errorMessage}");
             }
         }
 
-        public async Task CreateTableAsync<T>() where T : new()
+        public void Dispose()
```
dhei: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#dispose-and-disposeboolean

The C# Dispose pattern should implement 2 dispose methods:
(1) `public void Dispose()` which can be called by the consumer of the class to dispose resources.
(2) `protected virtual void Dispose(bool disposing)` which will be called by the both the consumer and the finalizer in gc.
This implementation doesn't handle the finalizer scenario.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
-                var task = (Task)_dbConnection.GetType()
-                    .GetMethod("CreateTableAsync", new[] { typeof(CreateFlags) })
-                    .MakeGenericMethod(typeof(T))
-                    .Invoke(_dbConnection, new object[] { CreateFlags.None });
-                await task.ConfigureAwait(false);
-            }
-            catch (SQLiteException e)
-            {
-                throw ToStorageException(e);
+                AppCenterLog.Error(AppCenterLog.LogTag, "Trying to dispose storage adapter while database is null.");
```
dhei: Per https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#dispose-and-disposeboolean,

> To help ensure that resources are always cleaned up appropriately, a Dispose method should be callable multiple times without throwing an exception.

Dispose pattern is supposed to be callable multiple times without side effects. We shouldn't log an error message if called a second time.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
-                return await table.Where(pred).Take(limit).ToListAsync().ConfigureAwait(false);
+                result = raw.sqlite3_bind_int(stmt, index, (int)value);
             }
-            catch (SQLiteException e)
+            else if (value is long)
             {
-                throw ToStorageException(e);
+                result = raw.sqlite3_bind_int64(stmt, index, (long)value);
+            }
+            else
```
dhei: No support for `sqlite3_bind_double` binding? 🤔

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
+                throw new StorageException($"Failed to bind {index} parameter, result={result}\n\t{errorMessage}");
             }
         }
 
-        public Task<int> CountAsync<T>(Expression<Func<T, bool>> pred) where T : new()
+        private void BindParameters(sqlite3_stmt stmt, IList<object> values)
         {
-            var table = _dbConnection.Table<T>();
-            return table.Where(pred).CountAsync();
+            for (int i = 0; i < values?.Count; i++)
```
dhei: `values?.Count` will return `null` when values is `null`, then comparing an integer with `null` will throw.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
+            {
+                BindParameter(stmt, i + 1, values[i]);
```
dhei: Let's add a comment to explain why the index is `i + 1` for each value. It's counter-intuitive.

https://www.sqlite.org/c3ref/bind_blob.html
> The second argument is the index of the SQL parameter to be set. The leftmost SQL parameter has an index of 1.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -238,8 +222,8 @@ public Task<string> GetLogsAsync(string channelName, int limit, List<Log> logs)
                     {
                         AppCenterLog.Error(AppCenterLog.LogTag, "Cannot deserialize a log in storage", e);
                         failedToDeserializeALog = true;
-                        _storageAdapter.DeleteAsync<LogEntry>(row => row.Id == entry.Id)
-                            .GetAwaiter().GetResult();
+                        var values = new List<object> { entry.Id };
```
dhei: Is `var values` used anywhere? I can't find it.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -276,12 +260,15 @@ private void ProcessLogIds(string channelName, string batchId, IEnumerable<Tuple
             AppCenterLog.Debug(AppCenterLog.LogTag, message);
         }
 
-        private async Task InitializeDatabaseAsync()
+        private void InitializeDatabase()
         {
+            Directory.CreateDirectory(Constants.AppCenterFilesDirectoryPath);
```
dhei: `Directory.CreateDirectory()` can throw IOException. Why not move it inside the try block?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -370,36 +357,31 @@ private Task<T> AddTaskToQueue<T>(Func<T> action)
...
             // But the message is definitely "Corrupt" and thus unfortunately that is the only check we seem to be able to do as opposed to type/property checking.
             if (e.Message == "Corrupt" || e.InnerException?.Message == "Corrupt")
             {
-                AppCenterLog.Error(AppCenterLog.LogTag, "Database corruption detected, deleting the file and starting fresh...", e);
-                await _storageAdapter.DeleteDatabaseFileAsync().ConfigureAwait(false);
-                await InitializeDatabaseAsync().ConfigureAwait(false);
+                AppCenterLog.Error(AppCenterLog.LogTag,
+                    "Database corruption detected, deleting the file and starting fresh...", e);
+                _storageAdapter.Dispose();
+                File.Delete(Constants.AppCenterDatabasePath);
```
dhei: `File.Delete` can throw `IOException`. It's not uncommon to hit the "file used by another process" error on Windows.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
+                var errorMessage = raw.sqlite3_errmsg(_db);
+                throw new StorageException($"Failed to create table, result={result}\n\t{errorMessage}");
             }
-            catch (SQLiteException e)
+        }
+
+        public int Count(string tableName, string columnName, object value)
+        {
+            var result = ExecuteSelectionSqlQuery($"SELECT COUNT(*) FROM {tableName} WHERE {columnName} = ?;", new[] { value });
+            return (int)(long)(result.FirstOrDefault()?.FirstOrDefault() ?? 0L);
```
dhei: Two typecast in a row is not very readable. Could you break it down into two lines?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
dhei: I was wrong. In C# comparing `null` with a number will result in `false`. So we are good.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
+        }
+
+        public IList<object[]> Select(string tableName, string columnName, object value, string excludeColumnName, object[] excludeValues, int? limit = null)
+        {
+            var whereClause = $"{columnName} = ?";
+            var args = new List<object> { value };
+            if (excludeValues?.Length > 0)
             {
-                throw ToStorageException(e);
+                whereClause += $" AND {excludeColumnName} NOT IN ({string.Join(",", Enumerable.Repeat("?", excludeValues.Length))})";
```
dhei: It's good practice not to make complex method calls inside C# string interpolation. It's less readable and harder to debug.

Let's make it more readable.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
-                }
-                if (_dbConnection == null)
-                {
-                    throw new StorageException("Cannot initialize SQLite library.");
-                }
-            });
+            var columnsClause = string.Join(",", columnNames);
+            var valueClause = string.Join(",", Enumerable.Repeat("?", values.First().Length));
+            var valuesClause = string.Join(",", Enumerable.Repeat($"({valueClause})", values.Count));
+            var valuesArray = values.SelectMany(i => i).ToArray();
```
dhei: Can you explain how does `values.SelectMany(i => i)` work? I don't understand how it would work.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,193 @@
...
-                    SQLiteAsyncConnection.ResetPool();
-                    var prefix = _databaseDirectory == null ? Constants.LocalAppData : "";
-                    new File(System.IO.Path.Combine(prefix, _databasePath)).Delete();
-                }
-                catch (Exception e)
-                {
-                    throw new StorageException(e);
-                }
-            });
+            var whereMask = $"{columnName} IN ({string.Join(",", Enumerable.Repeat("?", values.Length))})";
```
dhei: Same as https://github.com/microsoft/appcenter-sdk-dotnet/pull/1246/files#r354620634.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
MatkovIvan: `[[1,2],[3,4],[5,6]]` -> `[1,2,3,4,5,6]`

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
MatkovIvan: Like on Apple platforms we're implementing only used types

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
Jamminroot: That's a good point :) Was left undeleted after refactor

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -370,36 +358,38 @@ private Task<T> AddTaskToQueue<T>(Func<T> action)
...
-            // Return exception to re-throw.
             if (e is StorageException)
             {
-                // This is the expected case, storage adapter already wraps exception as StorageException, so return as is.
                 return e;
             }
-
-            // Tasks should already be throwing only storage exceptions, but in case any are missed, 
-            // which has happened (the Corrupt exception mentioned previously), catch them here and wrap in a storage exception. This will prevent 
-            // the exception from being unobserved.
```
MatkovIvan: Let's restore these comments

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -413,6 +403,7 @@ private void AddTaskToQueue(Task task)
             {
                 throw new StorageException("The operation has been canceled");
             }
+
```
MatkovIvan:
```suggestion
```

EatonZ:
Thanks for doing this, hope this will also open the door to strong-signing all your DLLs now? https://github.com/microsoft/appcenter-sdk-dotnet/issues/1092

MatkovIvan:
@EatonZ it's one of the reasons of this PR, yes

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -89,8 +82,11 @@ public Task PutLog(string channelName, Log log)
             return AddTaskToQueue(() =>
             {
                 var logJsonString = LogSerializer.Serialize(log);
-                var logEntry = new LogEntry { Channel = channelName, Log = logJsonString };
-                _storageAdapter.InsertAsync(logEntry).GetAwaiter().GetResult();
+                _storageAdapter.Insert(TableName,
+                    new[] { ColumnChannelName, ColumnLogName },
+                    new List<object[]> {
```
guperrot: do we actually insert non string values? we should change type in interface.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -109,20 +105,8 @@ public Task DeleteLogs(string channelName, string batchId)
...
-                    }
-                    AppCenterLog.Debug(AppCenterLog.LogTag, deletedIdsMessage);
-                    foreach (var id in identifiers)
-                    {
-                        _storageAdapter
-                            .DeleteAsync<LogEntry>(entry => entry.Channel == channelName && entry.Id == id)
-                            .GetAwaiter().GetResult();
-                    }
+                    AppCenterLog.Debug(AppCenterLog.LogTag, "The IDs for deleting log(s) is/are:\n\t" + string.Join("\n\t", identifiers));
+                    _storageAdapter.Delete(TableName, ColumnIdName, identifiers.Cast<object>().ToArray());
```
guperrot: still need to remove in memory pendingDbIdentifiers, why was the code removed?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -352,7 +339,7 @@ private Task AddTaskToQueue(Action action)
                 catch (Exception e)
                 {
                     // If we use await on a task that returns an exception, it makes the app crash, however, using the awaiter works...
```
guperrot: outdated comment

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -370,24 +357,33 @@ private Task<T> AddTaskToQueue<T>(Func<T> action)
                 catch (Exception e)
                 {
                     // And regarding the comment in other variant of this function, async lambda + await cannot work with a Func anyway.
```
guperrot: outdated comment

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -370,24 +357,33 @@ private Task<T> AddTaskToQueue<T>(Func<T> action)
...
             });
             AddTaskToQueue(task);
             return task;
         }
 
-        private async Task<Exception> HandleStorageRelatedExceptionAsync(Exception e)
+        private Exception HandleStorageRelatedExceptionAsync(Exception e)
         {
             // Check if database is corrupted, we have evidence (https://github.com/microsoft/appcenter-sdk-dotnet/issues/1184)
             // that it's not always originated by a proper SQLiteException (which would then be converted to StorageException in StorageAdapter).
```
guperrot: Is it still SQLiteException? The comment might be outdated

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -370,24 +357,33 @@ private Task<T> AddTaskToQueue<T>(Func<T> action)
...
             // If it was always the right type then the exception would not have been unobserved in that application before we changed the re-throw logic here.
             // But the message is definitely "Corrupt" and thus unfortunately that is the only check we seem to be able to do as opposed to type/property checking.
             if (e.Message == "Corrupt" || e.InnerException?.Message == "Corrupt")
```
guperrot: Is is still valid check with the raw lib?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -370,24 +357,33 @@ private Task<T> AddTaskToQueue<T>(Func<T> action)
...
             {
-                AppCenterLog.Error(AppCenterLog.LogTag, "Database corruption detected, deleting the file and starting fresh...", e);
-                await _storageAdapter.DeleteDatabaseFileAsync().ConfigureAwait(false);
-                await InitializeDatabaseAsync().ConfigureAwait(false);
+                AppCenterLog.Error(AppCenterLog.LogTag,
```
guperrot: Did you manually test this with the new lib? To see if we correctly start over when simulating a corrupt somewhere (we tested adhoc by throwing a corrupt exception in insert last time)

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,203 @@
...
+            }
         }
 
-        public async Task<int> DeleteAsync<T>(Expression<Func<T, bool>> pred) where T : new()
+        public int Count(string tableName, string columnName, object value)
         {
-            try
+            var result = ExecuteSelectionSqlQuery($"SELECT COUNT(*) FROM {tableName} WHERE {columnName} = ?;", new[] { value });
+            var count = (long)(result.FirstOrDefault()?.FirstOrDefault() ?? 0L);
+            return (int)count;
```
guperrot: Why not using long all the way through and modify interface?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,203 @@
...
+                whereClause += $" AND {excludeColumnName} NOT IN ({BuildBindingMask(excludeValues.Length)})";
+                args.AddRange(excludeValues);
             }
-            catch (SQLiteException e)
+            var limitClause = limit != null ? $" LIMIT {limit}" : string.Empty;
+            var query = $"SELECT * FROM {tableName} WHERE {whereClause}{limitClause};";
+            return ExecuteSelectionSqlQuery(query, args);
+        }
+
+        public void Insert(string tableName, string[] columnNames, ICollection<object[]> values)
```
guperrot: Looks like the only place we use it is a string. We should change the insert code all the way from interface to here I think.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -1,158 +1,203 @@
...
+            if (result != raw.SQLITE_OK)
             {
-                throw ToStorageException(e);
+                var errorMessage = raw.sqlite3_errmsg(_db);
+                throw new StorageException($"Failed to prepare insert SQL query, result={result}\n\t{errorMessage}");
             }
         }
 
-        public Task InitializeStorageAsync()
+        public void Delete(string tableName, string columnName, params object[] values)
```
guperrot: Do we actually pass objects that are not string?

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Microsoft.AppCenter.WindowsDesktop.csproj
```diff
@@ -14,16 +14,17 @@
...
     <PackageReference Include="System.Management" Version="4.6.0" />
     <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.6.0" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)'=='net45'">
     <Reference Include="Microsoft.CSharp" />
-    <Reference Include="System.Management" />
     <Reference Include="System.Configuration" />
     <Reference Include="System.Deployment" />
+    <Reference Include="System.Management" />
+    <Reference Include="System.Net.Http" />
```
guperrot: .net core is not impacted by missing http?

Tests/Microsoft.AppCenter.Test.Windows/packages.config
```diff
@@ -7,7 +7,6 @@
   <package id="MSTest.TestFramework" version="1.1.18" targetFramework="net452" />
   <package id="NETStandard.Library" version="1.6.1" targetFramework="net452" />
   <package id="Newtonsoft.Json" version="12.0.2" targetFramework="net452" />
-  <package id="sqlite-net-pcl" version="1.3.1" targetFramework="net452" />
   <package id="SQLitePCLRaw.bundle_green" version="1.1.2" targetFramework="net452" />
```
guperrot: this should be updated everywhere

nuget/AppCenter.nuspec
```diff
@@ -18,15 +18,15 @@
...
         <dependency id="Newtonsoft.Json" version="12.0.2"/>
-        <dependency id="sqlite-net-pcl" version="1.3.1"/>
+        <dependency id="SQLitePCLRaw.bundle_green" version="[1.1.14, 2.0.0)" />
       </group>
       <group targetFramework="netcoreapp3.0">
         <dependency id="Newtonsoft.Json" version="12.0.2"/>
-        <dependency id="sqlite-net-pcl" version="1.3.1"/>
+        <dependency id="SQLitePCLRaw.bundle_green" version="[1.1.14, 2.0.0)" />
         <dependency id="System.Management" version="4.6.0"/>
         <dependency id="System.Configuration.ConfigurationManager" version="4.6.0"/>
```
guperrot: Don't we need https://www.nuget.org/packages/System.Net.Http/ on .NET core? Can we test with internal nugets in demo to check?

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Microsoft.AppCenter.WindowsDesktop.csproj
guperrot: Why the need for that in the first place?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
dhei: This is really cool, thanks.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
MatkovIvan: Yes - integers.

> we should change type in interface.

??

MatkovIvan: To array of strings? No.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
MatkovIvan: We thought about it, but it leads to modify interface if storage too. Cast here to minimize the changes. Anyway, int more than enough here

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
MatkovIvan: Disagree, we can introduce other types (for priority for example) to match with other SDK, it will lead to revert of revert.
`string` in parameter here doesn't look like the right thing.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
MatkovIvan: Yes. Integers to filter by ids

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
MatkovIvan: https://github.com/microsoft/appcenter-sdk-dotnet/pull/1246#discussion_r355109886

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
MatkovIvan: Also, there are others methods here that requires this abstraction now. Let's keep it harmonized.

I don't think that "it isn't used at the moment" is the reason to drop already implemented correct abstraction.

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Microsoft.AppCenter.WindowsDesktop.csproj
MatkovIvan: It was already discussed: .net core doesn't seem affected, need - it fails without it (It was referenced implicitly before somehow).
We'll recheck it again.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -97,14 +97,18 @@ private object GetColumnValue(sqlite3_stmt stmt, int index)
         private int ExecuteNonSelectionSqlQuery(string query, IList<object> args = null)
         {
             var db = _db ?? throw new StorageException("The database wasn't initialized.");
-            var result = raw.sqlite3_prepare_v2(db, query, out var stmt);
-            if (result != raw.SQLITE_OK)
+            var prepareResult = raw.sqlite3_prepare_v2(db, query, out var stmt);
+            if (prepareResult == raw.SQLITE_CORRUPT || prepareResult == raw.SQLITE_NOTADB)
+            {
+                throw new StorageCorruptException();
```
MatkovIvan: Add error message

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -117,11 +121,15 @@ private List<object[]> ExecuteSelectionSqlQuery(string query, IList<object> args
         {
             var db = _db ?? throw new StorageException("The database wasn't initialized.");
             var entries = new List<object[]>();
-            var queryResult = raw.sqlite3_prepare_v2(db, query, out var stmt);
-            if (queryResult != raw.SQLITE_OK)
+            var prepareResult = raw.sqlite3_prepare_v2(db, query, out var stmt);
+            if (prepareResult == raw.SQLITE_CORRUPT || prepareResult == raw.SQLITE_NOTADB)
+            {
+                throw new StorageCorruptException();
```
MatkovIvan: Add error message

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -97,14 +97,18 @@ private object GetColumnValue(sqlite3_stmt stmt, int index)
         private int ExecuteNonSelectionSqlQuery(string query, IList<object> args = null)
         {
             var db = _db ?? throw new StorageException("The database wasn't initialized.");
-            var result = raw.sqlite3_prepare_v2(db, query, out var stmt);
-            if (result != raw.SQLITE_OK)
+            var prepareResult = raw.sqlite3_prepare_v2(db, query, out var stmt);
+            if (prepareResult == raw.SQLITE_CORRUPT || prepareResult == raw.SQLITE_NOTADB)
```
MatkovIvan: Add this check to initialize function

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageCorruptException.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.AppCenter.Storage
+{
+    internal class StorageCorruptException : AppCenterException
```
MatkovIvan: StorageCorruptedException : StorageException

nuget/AppCenter.nuspec
MatkovIvan: I guess it's still referenced implicitly (dependency of dependency) via `NETStandard.Library`

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Microsoft.AppCenter.WindowsDesktop.csproj
MatkovIvan: Also, it's not new topic. Some users were affected by this: https://github.com/microsoft/appcenter-sdk-dotnet/issues/1134#issuecomment-549751706

I don't think that it affects core build, but we should add a requirement of it for .net framework in nuspecs

MatkovIvan: Created a separate PR with the fix (to develop) since it is NOT related to this feature.
See #1251

nuget/AppCenter.nuspec
MatkovIvan: https://github.com/microsoft/appcenter-sdk-dotnet/pull/1246#discussion_r355437350

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
Jamminroot: Yeah, I've tested the new impementation today; Works as expected

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
```diff
@@ -190,9 +182,18 @@ public void Delete(string tableName, string columnName, params object[] values)
...
+                throw CreateStorageException(result, "Failed to prepare delete SQL query");
+            }
+        }
+        private StorageException CreateStorageException(int result, string message)
+        {
+            var errorMessage = raw.sqlite3_errmsg(_db);
+            var exceptionMessage = $"{message}, result={result}\n\t{errorMessage}";
+            if (result == raw.SQLITE_CORRUPT || result == raw.SQLITE_NOTADB)
+            {
+                throw new StorageCorruptedException(exceptionMessage);
```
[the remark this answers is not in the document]
MatkovIvan: Yeah, it should be return

Tests/Microsoft.AppCenter.Analytics.Test.Windows/app.config
```diff
@@ -0,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
```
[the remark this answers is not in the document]
MatkovIvan: It was generated by IDE when we updated all SQLite versions in tests. It's more careful update than before

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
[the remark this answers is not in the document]
MatkovIvan:
```suggestion
                return new StorageCorruptedException(exceptionMessage);
```

guperrot:
Should do another round of testing with nugets built from this branch in the demo apps.

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/StorageAdapter.cs
dhei: @Jamminroot 

> The Dispose method performs all object cleanup, so the garbage collector no longer needs to call the objects' Object.Finalize override. Therefore, the call to the SuppressFinalize method prevents the garbage collector from running the finalizer. 

Per C# docs, the best practice is to call `GC.SuppressFinalize(this)` in `public void Dispose()` to tell GC that no need to clean resources again.

```c#
public void Dispose()
{
    Dispose(true);
    GC.SuppressFinalize(this);
}
```

MatkovIvan: `SuppressFinalize` should only be called by a class that has a finalizer, no need for it in this case

dhei: @Jamminroot 
Ignore my previous comment.
`GC.SuppressFinalize` is only when finalizer is implemented in the class. Since `StorageAdapater` class doesn't implement a finalizer, `GC.SuppressFinalize` is not needed.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            Microsoft.AppCenter.Utils.Constants.AppCenterDatabasePath = DatabasePath;
+            try
+            {
+                adapter.Initialize(DatabasePath);
+            }
+            catch
+            {
+                // Handle exception, database is not created with Mock.
+            }
+            Assert.IsTrue(System.IO.File.Exists(DatabasePath));
```
dhei: To avoid other test cases interfering, it would be ideal to check `DatabasePath` doesn't exist before `adapter.Initialize` call.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            {
+                adapter.Initialize(DatabasePath);
+            }
+            catch (Exception e)
+            {
+                Assert.Fail("Shouldn't have thrown exception");
+            }
+            try
+            {
+                // Try get data before database initialize.
```
dhei: This comment doesn't make sense. Is it a copy-and-paste from other test cases?

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+                CreateTableHelper(adapter);
+            }
+            catch (Exception e)
+            {
+                Assert.Fail("Shouldn't have thrown exception");
+            }
+            adapter.Dispose();
+            try
+            {
+                // Try get data before database initialize.
```
dhei: This comment doesn't make sense. Is it a copy-and-paste from other test cases?

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            {
+                // Initialize database.
+                _adapter.Initialize(DatabasePath);
+            }
+            catch
+            {
+                // Handle exception, database is not created with Mock.
+            }
+            CreateTableHelper(_adapter);
+            try
```
dhei: What is the difficult outcome of your code and the following code when test executed in CI or automation?

```c#
_adapter.Count(TableName, ColumnChannelName, StorageTestChannelName);
```

At least in Visual Studio Text Explorer, both will fail, not much difference.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            {
+                _adapter.Count(TableName, $"{ColumnChannelName}", true);
+                Assert.Fail("Should have thrown exception");
+            }
+            catch (Exception e)
+            {
+                Assert.IsTrue(e.Message.Contains("not supported"));
+            }
+
```
dhei:
```suggestion
```

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            }
+            catch
+            {
+                // Handle exception, database is not created with Mock.
+            }
+            Assert.IsTrue(System.IO.File.Exists(DatabasePath));
+        }
+
+        [TestMethod]
+        public void FaildToOpenDatabaseWhenNameWrong()
```
dhei:
```suggestion
        public void FailedToOpenDatabaseWhenNameWrong()
```

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            {
+                Assert.IsTrue(e.Message.Contains("not supported"));
+            }
+            
+        }
 
-            // Verify that a directory object was not created.
-            Assert.IsNull(adapter._databaseDirectory);
+        [TestMethod]
+        public void FaildToPrepareDatabaseWhenDelete()
```
dhei:
```suggestion
        public void FailedToPrepareDatabaseWhenDelete()
```

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+        {
+            string whereClause = $"{ColumnChannelName} = 'faild-value'.";
```
dhei:
```suggestion
            string whereClause = $"{ColumnChannelName} = 'failed-value'.";
```

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            catch (Exception e)
+            {
+                Assert.IsTrue(e.Message.Contains("Failed to prepare SQL query"));
+            }
+        }
 
-            // Should not crash even if directory is null.
-            adapter.InitializeStorageAsync().Wait();
+        [TestMethod]
+        public void DatabaseIsNotInitilazeWhenCallDelete()
```
dhei:
```suggestion
        public void DatabaseIsNotInitilazeWhenDelete()
```

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
 
-            // Replace the directory with a mock and initialize.
-            adapter._databaseDirectory = Mock.Of<Directory>();
-            adapter.InitializeStorageAsync().Wait();
-            Mock.Get(adapter._databaseDirectory).Verify(directory => directory.Create());
+        /// <summary>
+        /// Verify that database is not initilaze.
+        /// </summary>
+        [TestMethod]
+        public void DatabaseIsNotInitilazeWhenCallCount()
```
dhei:
```suggestion
        public void DatabaseIsNotInitilazeWhenCount()
```

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            var entryId = 0L;
+            testEntries.ForEach(entry =>
+            {
+                entryId = (long)entry[0];
+                Assert.AreEqual(entry[1], StorageTestChannelName);
+                Assert.AreEqual(entry[2], "");
+            });
+            _adapter.Delete(TableName, ColumnIdName, new object[] { entryId });
+            count = _adapter.Count(TableName, ColumnChannelName, StorageTestChannelName);
+            Assert.AreEqual(count, 0);
```
dhei:
```suggestion
            Assert.AreEqual(0, count);
```
C# `Assert.AreEqual` first parameter is expected value, second is actual value.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+        {
+            string[] tables = new string[] { ColumnIdName, ColumnChannelName, ColumnLogName };
+            string[] types = new string[] { "INTEGER PRIMARY KEY AUTOINCREMENT", "TEXT NOT NULL", "TEXT NOT NULL" };
+            adapter.CreateTable(TableName, tables, types);
+        }
+
+        private void InsertToTableHelper(IStorageAdapter adapter)
+        {
+            adapter.Insert(TableName,
+            new[] { ColumnChannelName, ColumnLogName },
```
dhei: Indentation incorrect.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+            var adapter = new StorageAdapter();
+            Microsoft.AppCenter.Utils.Constants.AppCenterFilesDirectoryPath = Environment.CurrentDirectory;
+            Microsoft.AppCenter.Utils.Constants.AppCenterDatabasePath = DatabasePath;
+            try
+            {
+                adapter.Initialize(DatabasePath);
+            }
+            catch
+            {
+                // Handle exception, database is not created with Mock.
```
dhei: I don't understand what you mean "database is not created with Mock" here. The test is using real storage with `StorageAdapter` not mocks. `adapter.Initialize(DatabasePath);` doesn't throw. What exception to catch here?

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+        public void InitializeStorageTest()
+        {
+            _adapter = new StorageAdapter();
+        }
+
         [TestMethod]
         public void InitializeStorageCreatesStorageDirectory()
         {
-            var adapter = new StorageAdapter("path/to/database/file.db");
+            var adapter = new StorageAdapter();
```
dhei: What's the reason for creating a new adapter instance instead of using `_adapter` property here?

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageTest.cs
```diff
@@ -3,40 +3,52 @@
...
         private const string StorageTestChannelName = "storageTestChannelName";
-        private readonly Microsoft.AppCenter.Storage.Storage _storage = new Microsoft.AppCenter.Storage.Storage();
+        private IStorage _storage;
+
+        // Const for storage data.
+        private const string TableName = "LogEntry";
+        private const string ColumnChannelName = "Channel";
+        private const string ColumnLogName = "Log";
+        private const string ColumnIdName = "Id";
+        private string DatabasePath;
```
dhei: Move this property before the "// Const for storage data." comment.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+
+        // Const for storage data.
+        private const string StorageTestChannelName = "storageTestChannelName";
+        private const string TableName = "LogEntry";
+        private const string ColumnChannelName = "Channel";
+        private const string ColumnLogName = "Log";
+        private const string ColumnIdName = "Id";
+        private const string DatabasePath = "databaseAtRoot.db";
+
+        [TestInitialize]
```
dhei: It would be useful to have a test case level cleanup method to avoid side-effects of manipulating storage.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageTest.cs
```diff
@@ -301,19 +313,37 @@ public void ClearPendingState()
         [TestMethod]
         public void FailToGetALog()
         {
-            var invalidLogEntry = new LogEntry { Channel = StorageTestChannelName, Log = "good luck deserializing me!" };
-            using (var connection = new SQLiteConnection("Microsoft.AppCenter.Storage"))
+            // Prepare data.
+            StorageAdapter adapter = new StorageAdapter();
+            var database = "test.db";
```
dhei: Why not use `DatabasePath` property? The reason we used GUID as database name is to avoid potential cleanup issues.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageTest.cs
```diff
@@ -301,19 +313,37 @@ public void ClearPendingState()
...
+            adapter.Initialize(database);
+            var tables = new string[] { ColumnIdName, ColumnChannelName, ColumnLogName };
+            var types = new string[] { "INTEGER PRIMARY KEY AUTOINCREMENT", "TEXT NOT NULL", "TEXT NOT NULL" };
+            adapter.CreateTable(TableName, tables, types);
+            adapter.Insert(TableName, tables,
+            new List<object[]>
```
dhei: Fix indentation.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
dhei: Ignore my previous comment.

Cleanup is in https://github.com/microsoft/appcenter-sdk-dotnet/pull/1246/files#diff-2368ae65b69493af7b1d82d1d642e7d5R275.

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageAdapterTest.cs
```diff
@@ -5,63 +5,286 @@
...
+        {
             try
             {
-                adapter.InitializeStorageAsync().Wait();
+                _adapter.Delete(TableName, ColumnChannelName, new object[] { StorageTestChannelName });
+                _adapter.Dispose();
+                _adapter = null;
             }
-            catch (AggregateException ex)
+            catch (Exception ignore)
```
dhei:
```suggestion
            catch
```

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageTest.cs
```diff
@@ -301,19 +313,37 @@ public void ClearPendingState()
...
-                var batchId = _storage.GetLogsAsync(StorageTestChannelName, 4, logs).RunNotAsync();
-                var count = _storage.CountLogsAsync(StorageTestChannelName).RunNotAsync();
-                Assert.IsNull(batchId);
-                Assert.AreEqual(0, logs.Count);
-                Assert.AreEqual(0, count);
+                adapter.Dispose();
+                storage.Dispose();
+                System.IO.File.Delete(database);
+            }
+            catch (Exception exc)
```
dhei:
```suggestion
            catch
```

Tests/Microsoft.AppCenter.Test.Windows/Storage/StorageTest.cs
```diff
@@ -3,40 +3,52 @@
...
 using System.Linq.Expressions;
 using System.Threading.Tasks;
-using LogEntry = Microsoft.AppCenter.Storage.Storage.LogEntry;
 
 namespace Microsoft.AppCenter.Test
 {
     [TestClass]
     public class StorageTest
     {
         private const string StorageTestChannelName = "storageTestChannelName";
```
dhei: Move private const string together.

PR #1253, opened 2019-12-10
Replace string split by regex to avoid xamarin android issues

guperrot:

## Description

Try to avoid missing method exception on xamarin android with string split

## Related PRs or issues

[AB#73940](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/73940)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Android
 
-* **[Fix]** Fix `MissingMethodException` that could occur with some build configurations (bug introduced in version 2.5.0).
+* **[Fix]** Fix `MissingMethodException` that could occur with some build configurations (a bug introduced in version `2.5.0`).
 
 ___
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Version 2.6.3 （Under development）
 
+### AppCenter
+
+#### Android
+
+* **[Fix]** Fix `MissingMethodException` that could occur with some build configurations (bug introduced in version 2.5.0).
+
 ___
 
 ## Version 2.6.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 #### Android
 
-* **[Fix]** Fix `MissingMethodException` that could occur with some build configurations (a bug introduced in version `2.5.0`).
+* **[Fix]** Fix `MissingMethodException` of `String.Split()` with some build configurations (a bug introduced in version `2.5.0`).
 
 ___
 
--- a/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
@@ -3,7 +3,7 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 namespace Microsoft.AppCenter
@@ -18,6 +18,13 @@ namespace Microsoft.AppCenter
         const string TargetKeyName = "target";
         const string TargetKeyNameUpper = "Target";
         const string AppSecretKeyName = "appsecret";
+        const string SecretsPattern = @"(?:([^;=]+)=([^;=]+));?";
+
+#if NETSTANDARD
+        static readonly Regex _secretsRegex = new Regex(SecretsPattern);
+#else
+        static readonly Regex _secretsRegex = new Regex(SecretsPattern, RegexOptions.Compiled);
+#endif
 
         // Gets the first instance of an app sceret and/or target token corresponding to the given platform name, or returns the string 
         // as-is if no identifier can be found. Logs a message if no identifiers can be found.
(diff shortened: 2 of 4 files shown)


guperrot:
https://regex101.com/r/Mm9tBA/3

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -37,13 +44,15 @@ internal static string GetSecretAndTargetForPlatform(string secrets, string plat
...
             }
 
-            // Grouping by the name of the key.
-            var secretsGroup = secrets.Split(SecretDelimiter.ToCharArray())
-                                           .Select(value => value.Split(PlatformKeyValueDelimiter.ToCharArray()))
-                                           .GroupBy(p => p[0] ?? "");
+            // Iterate over matching patterns.
+            var secretsDictionary = new Dictionary<string, string>();
+            var matches = _secretsRegex.Matches(secrets);
+            foreach (Match match in matches)
```
MatkovIvan:
```suggestion
            foreach (var match in matches)
```

Tests/Microsoft.AppCenter.Test.Windows/AppCenterTest.cs
```diff
@@ -657,6 +657,32 @@ public void ConfigureAppCenterWithEmptyAppSecretEmptyResult()
...
+        {
+            var appSecret = Guid.NewGuid().ToString();
+            var platformId = "uwp";
+            var secrets = $"=;{platformId}={appSecret};=";
+            var parsedSecret = AppCenter.GetSecretAndTargetForPlatform(secrets, platformId);
+            Assert.AreEqual(appSecret, parsedSecret);
+        }
+
+        /// <summary>
+        /// Verify last value is used on duplicate key
```
MatkovIvan:
```suggestion
        /// Verify last value is used on duplicate key.
```

Tests/Microsoft.AppCenter.Test.Windows/AppCenterTest.cs
```diff
@@ -657,6 +657,32 @@ public void ConfigureAppCenterWithEmptyAppSecretEmptyResult()
             Assert.IsFalse(AppCenter.Configured);
         }
 
+        /// <summary>
+        /// Verify empty pairs are ignored
```
MatkovIvan:
```suggestion
        /// Verify empty pairs are ignored.
```

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Version 2.6.3 （Under development）
 
+### AppCenter
+
+#### Android
+
+* **[Fix]** Fix `MissingMethodException` that could occur with some build configurations (a bug introduced in version `2.5.0`).
```
dhei: Let's where the exception is thrown.

```suggestion
* **[Fix]** Fix `MissingMethodException` of `String.Split()` with some build configurations (a bug introduced in version `2.5.0`).
```

SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -18,6 +18,13 @@ public partial class AppCenter
         const string TargetKeyName = "target";
         const string TargetKeyNameUpper = "Target";
         const string AppSecretKeyName = "appsecret";
+        const string SecretsPattern = @"([^;=]+)=([^;=]+);?";
+
+#if NETSTANDARD
+        static readonly Regex _secretsRegex = new Regex(SecretsPattern);
+#else
+        static readonly Regex _secretsRegex = new Regex(SecretsPattern, RegexOptions.Compiled);
```
dhei: What is the reason to use `RegexOptions.Compiled` option when it's not NET Standard?

https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options?view=netframework-4.8#compiled-regular-expressions
> By default, regular expressions in .NET are interpreted. When a Regex object is instantiated or a static Regex method is called, the regular expression pattern is parsed into a set of custom opcodes, and an interpreter uses these opcodes to run the regular expression. This involves a tradeoff: The cost of initializing the regular expression engine is minimized at the expense of run-time performance.

Performance-wise in C#, comparing interpreted regex to compiled regex, interpreted regex saves time at initialization but less efficient at execution time. C# documentation recommends using interpreted regex (default) when used infrequently.

https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices#interpreted-vs-compiled-regular-expressions
> we recommend that you use interpreted regular expressions when you call regular expression methods with a specific regular expression relatively infrequently. You should use compiled regular expressions when you call regular expression methods with a specific regular expression relatively frequently.

In this case, regex is used to parse app secrets string which is only called once per app lifecycle which is very infrequent, therefore from a perf standpoint, interpreted regex would be a more reasonable choice.

guperrot: .NET Standard 1.0 does not have this option.
But the code never runs on .NET standard anyway, we bait and switch the real implementation.

dhei: The "expected" usage pattern of `AppCenter.Start` is to be called at app startup time once per lifecycle (infrequent), though there is nothing stops developers to call them on every `onResume` (frequent). The compiled only has advantage when the regex is called frequently which is not the "mainstream" way to consume this API.

From what I read online, the perf difference of interpreted regex and compiled regex in C# is probably not significant no matter which one we choose. But to know it for sure, the only way is to benchmark testing it using StopWatch--I don't think it's worth it. 

I slightly disagree but not a big deal.

guperrot: We document to add AppCenter.start in Xamarin Forms onStart() which happens multiple times per process on Android (for example back button and start again).

dhei: That's fair. Thanks.

PR #1288, opened 2020-01-14
Allow whitespaces between 2 application secrets

guperrot:

## Description

Allow whitespaces after `;` secret separator only between secrets (no whitespace allowed before first or after last if `;` missing).
Do not allow garbage character in between secrets.

## Related PRs or issues

[AB#75074](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75074)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### App Center
 
-* **[Feature]** Allow whitespaces such as newlines between 2 application secrets as long as the previous secret ends with `;`.
+* **[Improvement]** Allow whitespaces such as newlines between 2 application secrets as long as the previous secret ends with `;`.
 
 #### UWP
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### App Center
 
+* **[Feature]** Allow whitespaces such as newlines between 2 application secrets as long as the previous secret ends with `;`.
+
 #### UWP
 
 * **[Feature]** Support ARM64 architecture.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### App Center
 
-* **[Improvement]** Allow whitespaces such as newlines between 2 application secrets as long as the previous secret ends with `;`.
+* **[Fix]** Allow whitespaces between secrets like in versions before `2.6.0`.
 
 #### UWP
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### App Center
 
-* **[Fix]** Allow whitespaces between secrets like in versions before `2.6.0`.
+* **[Fix]** Allow whitespace characters between secrets like in versions before `2.6.0`.
 
 #### UWP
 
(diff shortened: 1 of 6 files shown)


SDK/AppCenter/Microsoft.AppCenter.Shared/AppCenter.cs
```diff
@@ -18,7 +18,7 @@ public partial class AppCenter
         const string TargetKeyName = "target";
         const string TargetKeyNameUpper = "Target";
         const string AppSecretKeyName = "appsecret";
-        const string SecretsPattern = @"([^;=]+)=([^;=]+);?";
+        const string SecretsPattern = @"([^;=]+)=([^;=]+)(;\s*)?";
```
MatkovIvan: 1. Why second group contains "not `=`"? Do we have the test for `id=123=123`?
2. In case (1), new line will be included into the first group of next match
3. We don't need to capture third group. It can be marked as non-captured (`?:`)

My suggestion is
```suggestion
        const string SecretsPattern = @"([^\s;=]+)=([^;]+)(?:;\s*)?";
```

guperrot: 1. 2. will try to improve the `=`.

3. we don't want to allow garbage characters with the suggestion you could pass:

`garbabe android=123;ios=456` and android would correctly parse.
with the one in the PR, android would not be recognized as a platform and trigger the error which according to me is better.

MatkovIvan:
```suggestion
        const string SecretsPattern = @"([^;=]+)=([^;]+)(?:;\s*)?";
```

MatkovIvan: 3. it's about do not capture third group

guperrot: I was talking about your first code suggestion that I disagreed. The new suggestion is ok.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### App Center
 
+* **[Fix]** Allow whitespaces between secrets like in versions before `2.6.0`.
```
michaelsmitha:
```suggestion
* **[Fix]** Allow whitespace between secrets like in versions before `2.6.0`.
```

Tests/Contoso.Test.Functional.Droid/Properties/AndroidManifest.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
dhei: Is this an invisible character change? Like end-of-line change or UTF-8 BOM encoding change?

guperrot: https://en.wikipedia.org/wiki/Byte_order_mark it comes and go because of windows vs mac vs where the last time update version script was run.

We could look into this with .gitattributes but not in this pr.

PR #1291, opened 2020-01-14
Analytics functional tests

AnastasiaKubova:

## Related PRs or issues

[AB#74719](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/74719)

### DIFF
--- a/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Android/Analytics.cs
+++ b/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Android/Analytics.cs
@@ -50,7 +50,7 @@ namespace Microsoft.AppCenter.Analytics
         /// <summary>
         /// Pause the Analytics service.
         /// </summary>
-        public static void Pause()
+        internal static void Pause()
         {
             AndroidAnalytics.Pause();
         }
@@ -58,7 +58,7 @@ namespace Microsoft.AppCenter.Analytics
         /// <summary>
         /// Resume the Analytics service.
         /// </summary>
-        public static void Resume()
+        internal static void Resume()
         {
             AndroidAnalytics.Resume();
         }
--- a/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Android/Analytics.cs
+++ b/SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Android/Analytics.cs
@@ -47,6 +47,22 @@ namespace Microsoft.AppCenter.Analytics
             return Task.Run(() => future.Get());
         }
 
+        /// <summary>
+        /// Pause the Analytics service.
+        /// </summary>
+        public static void Pause()
+        {
+            AndroidAnalytics.Pause();
+        }
+
+        /// <summary>
+        /// Resume the Analytics service.
+        /// </summary>
+        public static void Resume()
+        {
+            AndroidAnalytics.Resume();
+        }
+
         ///// <summary>
         ///// Enable or disable automatic page tracking.
         ///// Set this to false to if you plan to call <see cref="TrackPage"/> manually.
--- a/Tests/Contoso.Test.Functional.iOS/AppDelegate.cs
+++ b/Tests/Contoso.Test.Functional.iOS/AppDelegate.cs
@@ -4,7 +4,6 @@
 using System;
 using Foundation;
 using Microsoft.AppCenter.Test.Functional;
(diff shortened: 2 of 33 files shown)


Tests/Contoso.Test.Functional.iOS/AppDelegate.cs
```diff
@@ -19,6 +19,9 @@ public class AppDelegate : RunnerAppDelegate
     {
         private const string ResultChannelHost = "127.0.0.1";
 
+        private UIApplication UiApplication;
+        private NSDictionary LaunchOptions;
```
russelarms: These fields are not used.

Tests/Microsoft.AppCenter.Test.Functional/Analytics/TrackEventTest.cs
```diff
@@ -89,5 +90,108 @@ public async Task TrackEventWithPropertiesAsync()
...
+            DependencyConfiguration.HttpNetworkAdapter = httpNetworkAdapter;
+
+            // Start App Center.
+            AppCenter.UnsetInstance();
+            Analytics.UnsetInstance();
+            AppCenter.LogLevel = LogLevel.Verbose;
+            AppCenter.Start(_appSecret, typeof(Analytics));
+
+            // Pause Analytics module.
+            Analytics.Pause();
```
guperrot: There are 2 problems with this test.
First it adds an unexpected feature we need to document as we are now binding publicly Pause and Resume.
Second, if you remove pause and resume, the outcome would be exactly the same as we don't wait, the batching being 3s. We would first need to wait like 4s and verify nothing has been sent. Then resume and do the rest of the test.

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Android/Analytics.cs
```diff
@@ -47,6 +47,22 @@ public static Task SetEnabledAsync(bool enabled)
             return Task.Run(() => future.Get());
         }
 
+        /// <summary>
+        /// Pause the Analytics service.
+        /// </summary>
+        public static void Pause()
```
guperrot: This is a new API and thus a new feature.
The problem is that we don't have a Windows implementation for now and we would need to update 1DS and appcenter docs for it.

Let's make it internal for now until it's officially a feature. (internals are visible to test assembly)
```suggestion
        internal static void Pause()
```

SDK/AppCenterAnalytics/Microsoft.AppCenter.Analytics.Android/Analytics.cs
```diff
@@ -47,6 +47,22 @@ public static Task SetEnabledAsync(bool enabled)
...
+        {
+            AndroidAnalytics.Pause();
+        }
+
+        /// <summary>
+        /// Resume the Analytics service.
+        /// </summary>
+        public static void Resume()
```
guperrot:
```suggestion
        internal static void Resume()
```

Tests/Contoso.Test.Functional.iOS/AppDelegate.cs
dhei: Even if the field is used, I would name it `UI` instead of `Ui`.

Tests/Microsoft.AppCenter.Test.Functional/Analytics/TrackEventTest.cs
```diff
@@ -89,5 +90,116 @@ public async Task TrackEventWithPropertiesAsync()
...
+
+            // Test TrackEvent.
+            Analytics.TrackEvent("TrackEvent 1");
+            Task.WaitAny(httpNetworkAdapter.HttpResponseTask, Task.Delay(5000));
```
dhei: For readability, let's add a comment:
```
            // Wait for 5 seconds to allow batching happen (3 seconds), and verify nothing has been sent.
            Task.WaitAny(httpNetworkAdapter.HttpResponseTask, Task.Delay(5000));
            Assert.Equal(0, httpNetworkAdapter.CallCount);
```

dhei: Same for https://github.com/microsoft/appcenter-sdk-dotnet/pull/1291/files#diff-b4a12c3dfee704bc3e6d1460129a6f42R168.

PR #1306, opened 2020-01-24
Set in update track

AnastasiaKubova:

* [ ] ~Are tests passing locally?~
* [ ] ~Did you add unit tests if this modifies the Windows code?~

## Description

Adding the public vs private track modes.

## Related PRs or issues

[AB#75286](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75286)


SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Android/Distribute.cs
```diff
@@ -51,6 +51,30 @@ static void HandleUpdateAction(UpdateAction updateAction)
             }
         }
 
+        static void SetUpdateTrack(UpdateTrack updateTrack)
```
guperrot: We could just use int value of the enum, on Android it's just an int.

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Shared/UpdateTrack.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+namespace Microsoft.AppCenter.Distribute
```
guperrot:
```suggestion

namespace Microsoft.AppCenter.Distribute
```

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.iOS.Bindings/ApiDefinition.cs
```diff
@@ -43,6 +43,16 @@ interface MSDistribute
         [Static]
         [Export("notifyUpdateAction:")]
         void NotifyUpdateAction(MSUpdateAction action);
+
+        // + (void)setUpdateTrack:(MSUpdateTrack)updateTrack
+        [Static]
+        [Export("setUpdateTrack:")]
+        void SetUpdateTrack(MSUpdateTrack action);
```
dhei:
```suggestion
        void SetUpdateTrack(MSUpdateTrack updateTrack);
```

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Android/Distribute.cs
dhei: This doesn't handle the invalid track properly. If the caller passes an invalid track, the switch statement won't do anything because there is no matching case or default case for the switch. But we want to log an error on invalid value.

The android sdk handles invalid value and does logging properly. Either pass through the invalid value to native sdk, OR add a default case to the switch statement to handle invalid value.

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.iOS/Distribute.cs
```diff
@@ -100,6 +100,30 @@ static void HandleUpdateAction(UpdateAction updateAction)
             }
         }
 
+        static void SetUpdateTrack(UpdateTrack updateTrack)
+        {
+            switch (updateTrack)
```
dhei: Same issue as android SetUpdateTrack, see my comment https://github.com/microsoft/appcenter-sdk-dotnet/pull/1306/files#r370883579.

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Android/Distribute.cs
```diff
@@ -51,6 +51,30 @@ static void HandleUpdateAction(UpdateAction updateAction)
...
+                case UpdateTrack.Private:
+                    AndroidDistribute.UpdateTrack = 2;
+                    break;
+            }
+        }
+
+        static UpdateTrack GetUpdateTrack()
+        {
+            var updateTrackValue = AndroidDistribute.UpdateTrack;
+            if (updateTrackValue == 2)
```
dhei: In C#, you can [cast the integer to enum directly](https://docs.microsoft.com/en-us/dotnet/api/system.enum?view=netframework-4.8#performing-conversions):
```csharp
(UpdateTrack)updateTrackValue
```

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Android/Distribute.cs
guperrot: Just pass the int directly honestly, a c# enum can be cast to int

AnastasiaKubova: We already have the same implementation [here](https://github.com/microsoft/appcenter-sdk-dotnet/blob/bb4d3ed7ff3195f93397a8845210070e0cc83262/SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Android/Distribute.cs#L39) for `HandleUpdateAction(UpdateAction updateAction)` method. I guess we should have a similar implementation here too.

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Shared/UpdateTrack.cs
```diff
@@ -0,0 +1,20 @@
...
+using System;
+
+namespace Microsoft.AppCenter.Distribute
+{
+    public enum UpdateTrack
+    {
+        /// <summary>
+        /// Releases from the public group that don't require authentication.
+        /// </summary>
+        Public = 1,
```
jwallra: Any particular reason we use 1 and 2 here instead of 0 and 1?

guperrot: Yes it's to be future proof if we want to combine them as flags in a future version (for example public | private would mean we would login and fall back to public if login failed).

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Android/Distribute.cs
guperrot: Yes we should, though it's out of scope for this PR or feature. Could be a bug.

SDK/AppCenterDistribute/Microsoft.AppCenter.Distribute.Shared/UpdateTrack.cs
jwallra: Ah so it's really more like 1 << 1 and 1<< 2 if we were using C terms

MatkovIvan: @jwallra `1 << 0` and `1 << 1` 😉

scripts/configuration/ac-build-config.xml
```diff
@@ -4,6 +4,7 @@
         <sdkVersion>2.6.5-SNAPSHOT</sdkVersion>
         <iosVersion>2.5.4-37+0601f482c34c090ac221ac6c137fbedb6386806f</iosVersion>
         <androidVersion>2.5.2-20+64aa3683e</androidVersion>
```
guperrot: Duplicate XML tag
```suggestion
```

PR #1318, opened 2020-02-04
Add an option to rollback signal remove/install to old method

guperrot:

## Description

This property set to false was tested in a prerelease on a real world application that was experiencing crashes caused by the AppCenter SDK. Setting it to false solved the crashes.

## Related PRs or issues

[AB#75989](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/75989)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@ App Center Data is [retired](https://aka.ms/MBaaS-retirement-blog-post) and has 
 #### iOS
 
 * **[Improvement]** Add information logs when signals are removed/restored during initialization.
-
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -36,7 +36,7 @@ App Center Data is [retired](https://aka.ms/MBaaS-retirement-blog-post) and has
 #### iOS
 
 * **[Improvement]** Add information logs when signals are removed/restored during initialization.
-
+* **[Feature]** Add `Crashes.UseMonoRuntimeSignalMethods` boolean property to call before `AppCenter.Start`, if set to `false` it forces the [old signal behavior](https://www.mono-project.com/docs/advanced/signals/#incomplete-solution) even if the runtime Mono version is 4.8 or higher. The default value is `true`, which means using the [current signal behavior](https://www.mono-project.com/docs/advanced/signals/#complete-solution) if the Mono version is 4.8 or higher (and falling back to the old one on older Mono versions). This is introduced in case the current signal behavior causes unexpected segmentation fault crashes.
```
MatkovIvan: What the minimal version of mono required now? I thought it's higher than 4.8.

My thought here is it can be because of reflection. Also, we still don't know what exactly the case is, so I'm not sure that new public API is OK at this stage

guperrot: We don't really have a choice for now.

guperrot: We can always deprecate the property if we can get a root cause later, we agreed to use that workaround in the meantime in standup anyway to unblock them.

guperrot: The docs will only be in troubleshooting page also, that was the other point discussed.

PR #1351, opened 2020-03-10
Access to path is denied on storage initialization

AnastasiaKubova:

## Related PRs or issues

[AB#77597](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/77597)

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
@@ -273,18 +273,24 @@ namespace Microsoft.AppCenter.Storage
             var databaseDirectoryPath = Path.GetDirectoryName(_databasePath);
             if (databaseDirectoryPath != string.Empty)
             {
-                var databaseDirectory = new Utils.Files.Directory(databaseDirectoryPath);
-                 if (databaseDirectory != null)
+                using (var databaseDirectory = new Utils.Files.Directory(databaseDirectoryPath))
                 {
-                    try
+                    if (databaseDirectory != null)
                     {
-                        databaseDirectory.Create();
+                        try
+                        {
+                            databaseDirectory.Create();
+                        }
+                        catch (Exception e)
+                        {
+                            throw new StorageException("Failed to create database directory.", e);
+                        }
                     }
-                    catch (Exception e)
+                    else
                     {
-                        throw new StorageException("Failed to create database directory.", e);
+                        AppCenterLog.Warn(AppCenterLog.LogTag, "The database direcory path is empty.");
                     }
-                } 
+                }
             }
            
             try
--- a/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
@@ -12,6 +12,7 @@ using Microsoft.AppCenter.Ingestion.Models;
 using Microsoft.AppCenter.Ingestion.Models.Serialization;
 using Microsoft.AppCenter.Utils;
 using Newtonsoft.Json;
+using FileUtils = Microsoft.AppCenter.Utils.Files;
 
 namespace Microsoft.AppCenter.Storage
 {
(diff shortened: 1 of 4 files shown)


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -269,9 +270,25 @@ private void ProcessLogIds(string channelName, string batchId, IEnumerable<Tuple
...
+                    try
+                    {
+                        databaseDirectory.Create();
+                    }
+                    catch (Exception e)
+                    {
+                        throw new StorageException("Failed to create database directory.", e);
+                    }
+                } 
+            }
```
MatkovIvan: else { log }

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -269,9 +270,25 @@ private void ProcessLogIds(string channelName, string batchId, IEnumerable<Tuple
 
         private void InitializeDatabase()
         {
+            var databaseDirectoryPath = Path.GetDirectoryName(_databasePath);
+            if (databaseDirectoryPath != string.Empty)
```
MatkovIvan: Do the same on delete db file

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -269,9 +270,31 @@ private void ProcessLogIds(string channelName, string batchId, IEnumerable<Tuple
...
+                {
+                    if (databaseDirectory != null)
+                    {
+                        try
+                        {
+                            databaseDirectory.Create();
+                        }
+                        catch (Exception e)
+                        {
+                            throw new StorageException("Failed to create database directory.", e);
```
MatkovIvan: Log instead of re-throw

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Utils/Files/Directory.cs
```diff
@@ -58,5 +59,9 @@ public virtual void Refresh()
         {
             _underlyingDirectoryInfo.Refresh();
         }
+
+        public void Dispose()
```
MatkovIvan: Why do you need it?

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -269,9 +270,31 @@ private void ProcessLogIds(string channelName, string batchId, IEnumerable<Tuple
 
         private void InitializeDatabase()
         {
+            var databaseDirectoryPath = Path.GetDirectoryName(_databasePath);
+            if (databaseDirectoryPath != string.Empty)
```
MatkovIvan: It's never empty, you haven't touch the absolute path in consts

Tests/Microsoft.AppCenter.Test.UWP/AppCenterTest.cs
```diff
@@ -108,6 +108,42 @@ public void TestCorrectDatabasePathPassedFromStorageInitialization()
             }
         }
 
+        [TestMethod]
+        public void TestDbInitializationFolderNotExists()
+        {
+            // Make sure database directory does not exist before test.
+            var databaseDirectory = new Microsoft.AppCenter.Utils.Files.Directory(Constants.AppCenterDatabasePath);
```
MatkovIvan: You can use system API directly in tests

Tests/Microsoft.AppCenter.Test.UWP/AppCenterTest.cs
```diff
@@ -108,6 +108,42 @@ public void TestCorrectDatabasePathPassedFromStorageInitialization()
...
+                // Verify that database is created inside local app data folder, and not locally.
+                Assert.IsTrue(File.Exists(Path.Combine(Constants.LocalAppData, Constants.AppCenterDatabaseFilename)));
+            }
+        }
+
+        [TestMethod]
+        public void TestDbInitializationFolderExists()
+        {
+            // Make sure database directory exists before test.
+            var databaseDirectory = new Microsoft.AppCenter.Utils.Files.Directory(Constants.AppCenterDatabasePath);
```
MatkovIvan: Same

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Storage/Storage.cs
```diff
@@ -283,6 +286,27 @@ private void InitializeDatabase()
...
+            {
+                var databaseDirectory = new Directory(databaseDirectoryPath);
+                if (!databaseDirectory.Exists())
+                {
+                    databaseDirectory.Create();
+                }
+            }
+            catch (Exception e)
+            {
+                AppCenterLog.Error(AppCenterLog.LogTag, "Failed to create database directory.", e);
```
guperrot: On second thought why are we not throwing this one and handle the exception in the existing function?
If we catch it here and directory failed to be created, then storage initialization will also fail, so 2 exceptions instead of one.

MatkovIvan: it can fails on try to access the directory but open the db file just fine. "Exist" check fixes it in this case, but anyway it does not mean that storage init will fail

guperrot: This is a very weird scenario then, can we add comments?

Jamminroot: Added comment

Issue #1088, opened 2019-08-16
WPF version not reported correctly when using ClickOnce


moshegutman:
### **Description**

It looks like the WPF analytics module is using the Assembly file version for "version" when viewing data in App Center. I would like it to use the ClickOnce publish version, or better yet, let me set the version myself.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Publish any WPF app with ClickOnce
2. The version number displayed on the charts does not reflect the ClickOnce publish version number.

Copied from
https://github.com/microsoft/appcenter/issues/907

AnastasiaKubova:
Hi @moshegutman! It looks like the feature request. Could you please clarify why change the version in the file which we use don't suitable for you?

moshegutman:
The version that is used for display in App Center does not match the ClickOnce publish version highlighted below. I think this is a bug, not a feature request. Every time a new version of my app is published, the ClickOnce version is incremented, but App Center always displays the same version number. This is because App Center uses the Assembly File Version instead.

![clickonce](https://user-images.githubusercontent.com/3697619/63282785-33f27280-c275-11e9-8924-056117bc9fcb.png)

lucahost:
Maybe this could have a relation to the appcenter distribute issue I experienced. 
I posted it to the dev-community https://developercommunity.visualstudio.com/idea/702256/error-deploying-msi-to-appcenter.html

annakocheshkova:
Hi there. We use the version from `Package.appxmanifest`. I am not familiar with **ClickOnce** technology but it looks like a custom thing. I will mark this as feature request and file a ticket.
Thank you.

moshegutman:
ClickOnce is not “custom” https://docs.microsoft.com/en-us/visualstudio/deployment/clickonce-security-and-deployment?view=vs-2019

It is a feature of Visual Studio. This is a bug, not a feature request.

annakocheshkova:
@moshegutman we will look into this in any case and fix. Thanks for bringing it into attention!

guperrot:
Rather than manually setting version, it should be better to try to use something like:
```
if (ApplicationDeployment.IsNetworkDeployed)
        return ApplicationDeployment.CurrentDeployment.CurrentVersion;
```

moshegutman:
Oh I agree, but I don’t know how or where App center pulls version information from.

guperrot:
[...]

PR #1365, opened 2020-04-10
Fix reporting WPF versions

MatkovIvan:

## Description

Currently, SDK incorrectly uses the same version for both fields. "appBuild" field should contains exact number of specific build to improve grouping of problems on portal.

Some info about difference between versions:

`AssemblyInformationalVersion` - the product version of the assembly. This is the version you would use when talking to customers or for display on your website. This version can be a string, like '1.0 Release Candidate'. Optional. If not given, the `AssemblyFileVersion` is used.

`AssemblyFileVersion` is intended to uniquely identify a build. It can be increased for every deployment/build.

Also see [a good descriptions on StackOverflow](https://stackoverflow.com/questions/64602).

## Related PRs or issues

[AB#79013](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/79013)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Version 3.1.1 (Under development)
 
+### App Center
+
+#### WPF
+
+* **[Fix]** Fix app build version reporting, the assembly file version is used for this field now.
+
 ### App Center Crashes
 
 * **[Fix]** Remove the multiple attachments warning as that is now supported by the portal.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### App Center
 
-#### WPF
+#### WPF/WinForms
 
 * **[Fix]** Fix app build version reporting, the assembly file version is used for this field now.
 
--- a/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
@@ -118,8 +118,14 @@ namespace Microsoft.AppCenter.Utils
 #endif
             // The AssemblyFileVersion uniquely identifys a build.
             var entryAssembly = Assembly.GetEntryAssembly();
-            var fileVersionInfo = FileVersionInfo.GetVersionInfo(entryAssembly.Location);
-            return fileVersionInfo.FileVersion;
+            if (entryAssembly != null)
+            {
+                var fileVersionInfo = FileVersionInfo.GetVersionInfo(entryAssembly.Location);
+                return fileVersionInfo.FileVersion;
+            }
+
+            // Fallback if entry assembly is not found (in unit tests for example).
+            return GetAppVersion();
         }
 
         protected override string GetScreenSize()
--- a/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
@@ -101,31 +101,16 @@ namespace Microsoft.AppCenter.Utils
         protected override string GetAppVersion()
         {
             /*
+             * Application.ProductVersion returns the value from AssemblyInformationalVersion.
              * If the AssemblyInformationalVersion is not applied to an assembly,
              * the version number specified by the AssemblyFileVersion attribute is used instead.
              */
-            return Application.ProductVersion;
+            return DeploymentVersion ?? Application.ProductVersion;
         }
 
         protected override string GetAppBuild()
         {
-#if NET461
-            // Get ClickOnce version (does not exist on .NET Core). 
-            if (ApplicationDeployment.IsNetworkDeployed)
-            {
-                return ApplicationDeployment.CurrentDeployment.CurrentVersion.ToString();
-            }
-#endif
-            // The AssemblyFileVersion uniquely identifys a build.
-            var entryAssembly = Assembly.GetEntryAssembly();
-            if (entryAssembly != null)
-            {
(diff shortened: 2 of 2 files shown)


EatonZ:
Hi @MatkovIvan, it looks like this will address https://github.com/microsoft/appcenter/issues/1313, but for my use case, it needs changing. If, with your changes and the versionioning setup I described in https://github.com/microsoft/appcenter/issues/1313, logging an exception to App Center looks like this:



For things to work right for me, you actually need to swap the version & build methods so the version in my screenshot will read: **3.0.0.19546 (2020.04.12.02.43.23)**

It's possible this may not work for all your customers, which was why I suggested adding the ability  to override the version & build so that we can provide the versions & builds that we actually want reported.

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
```diff
@@ -95,19 +100,26 @@ protected override string GetOsVersion()
...
-            if (System.Deployment.Application.ApplicationDeployment.IsNetworkDeployed)
-            {
-                return System.Deployment.Application.ApplicationDeployment.CurrentDeployment.CurrentVersion.ToString();
-            }
-#endif
+            /*
+             * If the AssemblyInformationalVersion is not applied to an assembly,
+             * the version number specified by the AssemblyFileVersion attribute is used instead.
+             */
             return Application.ProductVersion;
```
guperrot, line 107: Why are we removing the ClickOnce support for the main version field?

MatkovIvan: That's why It's marked as "do not merge" - it's under discussion now. Currently, I changed versions in following way:
App Version - human readable
App Build - unique build/deployment

guperrot: This will break #1088 use case as Analytics uses only appVersion. We cannot make such a risky change right now.

I agree appBuild needs to be different than appVersion when ClickOnce is not used (the rest of the patch) but I think for ClickOnce we need to use it for both to avoid breaking scenarios.

MatkovIvan:
@EatonZ thanks for the comments. Yes, this one is partially solves your issue, but public setters is a different one. It will be addressed separately

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Version 3.1.1 (Under development)
 
+### App Center
+
+#### WPF
```
guperrot, line 7: I thought WinForms too?

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
```diff
@@ -123,6 +126,38 @@ protected override string GetScreenSize()
...
+                // The AssemblyFileVersion uniquely identifys a build.
+                var entryAssembly = Assembly.GetEntryAssembly();
+                if (entryAssembly != null)
+                {
+                    var fileVersionInfo = FileVersionInfo.GetVersionInfo(entryAssembly.Location);
+                    return fileVersionInfo.FileVersion;
+                }
+
+                // Fallback if entry assembly is not found (in unit tests for example).
+                return Application.ProductVersion;
```
guperrot, line 157: Shouldn't it be `DeploymentVersion ?? Application.ProductVersion` fallback?

MatkovIvan: No, it's already a fallback if `DeploymentVersion` is not available

CHANGELOG.md
MatkovIvan:
```suggestion
#### WPF/WinForms
```

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
guperrot, line 157: But the way the code is written, it would make more sense to return null here and use ?? in the caller

MatkovIvan: Initially I thought that it makes more sense, but if take into account why it's here and what this fallback really is, then it's better to keep it as is.
This line is mainly for unit tests (or any unknown cases), where entry assembly is not available. It will be just more confusing if move this fallback to caller since it's not supposed to be called in regular situation. The line already has a comment about it.

Let's keep this fallback here to remain code in main function clear.

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
```diff
@@ -123,6 +126,38 @@ protected override string GetScreenSize()
...
+#endif
+                return null;
+            }
+        }
+
+        private static string FileVersion
+        {
+            get
+            {
+                // The AssemblyFileVersion uniquely identifys a build.
```
annakocheshkova, line 148: nit
```suggestion
                // The AssemblyFileVersion uniquely identifies a build.
```

TranstecJasonD:
Unless I misunderstand, AssemblyFileVersion should not be used for reporting. At least I certainly do not want it to be. My app is composed of an EXE and several DLLs. Sometimes I when I do a build, only the DLLs are changed so the AssemblyFileVersion is not changed on the EXE because it did not change. I increment the AssemblyInformationalVersion instead, because it's a new product version, not a new EXE.

MatkovIvan:
@TranstecJasonD `AssemblyInformationalVersion` will be used if exists. `AssemblyFileVersion` is just a fallback

PR #1391, opened 2020-05-14
Xamarin functional tests timeout fix

olmisa:

* [ ] ~Did you add unit tests if this modifies the Windows code?~
* [ ] ~Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?~

## Description

Sometimes distribute tests were failed on CI with timeout. Excess mocked response json was removed.

## Related PRs or issues

[AB#79781](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/79781)

### DIFF
--- a/Tests/Microsoft.AppCenter.Test.Functional/Distribute/DistributeUpdateTest.cs
+++ b/Tests/Microsoft.AppCenter.Test.Functional/Distribute/DistributeUpdateTest.cs
@@ -173,8 +173,9 @@ namespace Microsoft.AppCenter.Test.Functional.Distribute
                 Content = GetReleaseJson("30", "3.0.0", false, 19),
                 StatusCode = 200
             };
+            var implicitCheckForUpdateTask = httpNetworkAdapter.MockRequest(request => request.Method == "GET" && request.Uri.Contains(urlDiff));
+            var explicitCheckForUpdateTask = httpNetworkAdapter.MockRequest(request => request.Method == "GET" && request.Uri.Contains(urlDiff), response);
             var startServiceTask = httpNetworkAdapter.MockRequestByLogType("startService");
-            var implicitCheckForUpdateTask = httpNetworkAdapter.MockRequest(request => request.Method == "GET");
 
             // Start AppCenter.
             AppCenter.UnsetInstance();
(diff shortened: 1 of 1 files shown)


Tests/Microsoft.AppCenter.Test.Functional/Distribute/DistributeUpdateTest.cs
```diff
@@ -197,23 +191,27 @@ public async Task CheckForUpdateTest(UpdateTrack updateTrack, string urlDiff)
...
             // Verify response.
-            Assert.Equal("GET", result.Method);
-            Assert.True(result.Uri.Contains(urlDiff));
-            Assert.True(result.Uri.Contains(Config.ResolveAppSecret()));
+            Assert.Equal("GET", resultImplicit.Method);
+            Assert.Contains(urlDiff, resultImplicit.Uri);
+            Assert.Contains(Config.ResolveAppSecret(), resultImplicit.Uri);
+
+            // Wait a 5s for give time to complete internal processes.
+            await Task.Delay(5000);
```
Jamminroot, line 202: This should work, but I don't think this fix is correct

Jamminroot: This is the initial implementation we had (we had this and something like `Task.WhenAny`...).

AnastasiaKubova: This way to avoid failing tests, because some internal operations need time to complete.
In failing cases, I found this log that appears after call `CheckForUpdate`:
`A check for update is already ongoing.`
And I suspect that [this](https://github.com/microsoft/appcenter-sdk-android/blob/develop/sdk/appcenter-distribute/src/main/java/com/microsoft/appcenter/distribute/Distribute.java#L499) returns false because `mCheckReleaseCallId` is still not null.

Jamminroot: I see. Ok, resolving this PR. Thanks for explanation o/

Issue #1398, opened 2020-06-10
Access report ID


abdelrahman-alfar:
**Is your feature request related to a problem? Please describe.**
We try to track the reportId to search easier on the appcenter when handling the exception we show a message box to our customers and we would like to include this reportID in the message.

at the begining tried to handle it using this event **Crashes.SentErrorReport** 
  private void Crashes_SentErrorReport(object sender, SentErrorReportEventArgs e)
        {
            var issueKey = e.Report.Id;
            this.showMessageService.ShowInfoBlocking(ErrorWindowStrings.PublishError, ErrorWindowStrings.SendErrorSuccessFormat(issueKey));
        }
but found out that this event is only triggered when the app crashes not when its handled

**Describe the solution you'd like**
You can return the report Id using   Crashes.TrackError(exception, properties, attachments) for example instead if its void

AnastasiaKubova:
Hi! Thanks for getting in touch with us! We will create a feature request for this.

DmitriyKirakosyan:
Hi @abdelrahman-alfar ,

I would like to refresh the status of this feature request with you. Could you please confirm that this feature request is still relevant for you? If yes, please provide some information on how exactly you would like to use this reportId and if there are any alternatives/workaround which you are currently using instead.

abdelrahman-alfar:
Hi @DmitriyKirakosyan 
I opened a pull request with it last year :) https://github.com/microsoft/appcenter-sdk-dotnet/pull/1401

DmitriyKirakosyan:
Hi @abdelrahman-alfar , yep we are aware of your PR and that's why I'm reaching out to you. Sorry for being unclear in my message.
First of all, thank you so much for your contribution! I would like to know more about your use case to have a better understanding how much value it can bring to SDK users.
In case we decide to merge the changes in your PR, we would also need to update all other platforms (android native, apple native, xamarin, react-native, unity) for the sake of consistency. That is why I would like to know more about how this feature is going to be used.

DmitriyKirakosyan:
Hi @abdelrahman-alfar , just a friendly ping.
Btw, as for searching a report for a specific user you can use an `userId` or a `sessionId`.

DmitriyKirakosyan:
I'm going to close this feature request but feel free to reopen if anyone is still interested in it.
Will also close the PR for now.

PR #1401, opened 2020-06-22
Added ErrorReportID access in Crashes #1398

abdelrahman-alfar:

## Description

A few sentences describing the overall goals of the pull request.

## Related PRs or issues

List related PRs and other issues.

## Misc

Add what's missing, notes on what you tested, additional thoughts or questions.

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
@@ -100,7 +100,7 @@ namespace Microsoft.AppCenter.Crashes
             Instance.InstanceHandlerUserConfirmation(userConfirmation);
         }
 
-        private static Guid PlatformTrackError(System.Exception exception, IDictionary<string, string> properties, ErrorAttachmentLog[] attachments)
+        private static Guid? PlatformTrackError(System.Exception exception, IDictionary<string, string> properties, ErrorAttachmentLog[] attachments)
         {
             return Instance.InstanceTrackError(exception, properties, attachments);
         }
@@ -378,13 +378,13 @@ namespace Microsoft.AppCenter.Crashes
             return Task.WhenAll(tasks);
         }
 
-        private Guid InstanceTrackError(System.Exception exception, IDictionary<string, string> properties, ErrorAttachmentLog[] attachments)
+        private Guid? InstanceTrackError(System.Exception exception, IDictionary<string, string> properties, ErrorAttachmentLog[] attachments)
         {
             lock (_serviceLock)
             {
                 if (IsInactive)
                 {
-                    return Guid.Empty;
+                    return null;
                 }
                 properties = PropertyValidator.ValidateProperties(properties, "HandledError");
                 var exceptionAndBinaries = ErrorLogHelper.CreateModelExceptionAndBinaries(exception);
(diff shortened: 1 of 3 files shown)


abdelrahman-alfar:
this is the implementation for https://github.com/microsoft/appcenter-sdk-dotnet/issues/1398

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -483,5 +486,7 @@ internal static void PlatformUnsetInstance()
         {
             Instance = null;
         }
+
+        public string ErrorReportID { get; set; }
```
guperrot, line 490: The ErrorReport object that is exposed in crash callbacks already has an Id property.

A static property here would not expose handled error ID but only the first pending crash from last run so it duplicates an existing feature and is not solving the feature request which is about handled errors.

What needs to be done instead is a modification on trackError instead to return the ID instead of void.

abdelrahman-alfar: updated it now track trace returns a guid and andusted the method to return errorId or empty Guid

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -483,5 +484,6 @@ internal static void PlatformUnsetInstance()
         {
             Instance = null;
         }
+
```
guperrot, line 487:
```suggestion
```

abdelrahman-alfar: Done Thanks

SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Windows.Shared/Crashes.cs
```diff
@@ -100,9 +100,9 @@ private static void PlatformNotifyUserConfirmation(UserConfirmation userConfirma
             Instance.InstanceHandlerUserConfirmation(userConfirmation);
         }
 
-        private static void PlatformTrackError(System.Exception exception, IDictionary<string, string> properties, ErrorAttachmentLog[] attachments)
+        private static Guid PlatformTrackError(System.Exception exception, IDictionary<string, string> properties, ErrorAttachmentLog[] attachments)
```
MatkovIvan, line 103: I see that only `private` methods were modified, so it's not clear how exactly this will solve the initial problem.

Probably, `SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.Shared/Crashes.cs` should be modified as well.

abdelrahman-alfar: pushed a newer commit, with those changes and also replaced GUID to be nullable to return null instead of guid.empty incase its inactive

abdelrahman-alfar:
the modification on ios and Android will be done from your side or i have to implement it for you?

MatkovIvan:
> Also, please note that we need to keep public method signatures equal between platforms to make possible usage from portable libraries

this PR should sync public signatures at least, the actual implementation for these platforms is out of scope

Ghoste87978 (approved):

DmitriyKirakosyan:
I'm closing this PR. Please upvote https://github.com/microsoft/appcenter-sdk-dotnet/issues/1398 if you're still interested in it.

PR #1471, opened 2020-11-18
Set max storage size API

olmisa:
Things to consider before you submit the PR:

* [x] Has `CHANGELOG.md` been updated?
* [ ] Are tests passing locally?
* [x] Are the files formatted correctly?
* [ ] Did you add unit tests if this modifies the Windows code?
* [x] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

`SetMaxStorageSize` API was added which allows setting a maximum size limit on the local SQLite storage.

## Related PRs or issues

[AB#83806](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/83806)

PR #1485, opened 2020-12-24
Start new 4.1.0 version

xabldint:


CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
...
+
+ ### App Center Crashes
+
+ #### Android
+
+* **[Fix]** Fix removing throwable files after rewriting error logs due to small database size.
+
+#### iOS
+
+* **[Fix]** Fix nullability for attachments methods in crashes delegate. `nil` is valid return value now (4.0.0 regression).
```
MatkovIvan, line 36:
```suggestion
```

MatkovIvan (changes requested):

CHANGELOG.md
LuyaLiu: ok, I've updated it.

Apps/Contoso.Android.Puppet/ModulePages/AppCenterFragment.cs
```diff
@@ -130,5 +142,16 @@ private void WriteLog(object sender, EventArgs e)
             string tag = LogWriteTagText.Text;
             LogFunctions[mLogWriteLevel](tag, message);
         }
+
+        private void SaveStorageSize(object sender, EventArgs e)
+        {
+            var inputText = StorageSizeText.Text;
+            var size = string.IsNullOrEmpty(inputText) ? 0 : long.Parse(inputText);
```
andrii-z4i, line 149: so, if inputText is not a number then we can easily get an exception here.
Not so safe approach. https://docs.microsoft.com/en-us/dotnet/api/system.int64.parse?view=net-5.0

Apps/Contoso.iOS.Puppet/ModulePages/AppCenterController.cs
```diff
@@ -83,5 +91,16 @@ public override void PrepareForSegue(UIStoryboardSegue segue, NSObject sender)
             string tag = LogWriteTag.Text;
             LogFunctions[mLogWriteLevel](tag, message);
         }
+
+        partial void SaveStorageSize(NSObject sender)
+        {
+            var inputText = StorageSizeText.Text;
+            if (int.TryParse(inputText, out var size))
```
andrii-z4i, line 98: and if we didn't parse text successfully what will happen? Any notification to user or log line?

Apps/Contoso.Android.Puppet/ModulePages/AppCenterFragment.cs
MatkovIvan: Hm. I thought we fixed all similar places based on my review some time ago - https://github.com/microsoft/appcenter-sdk-dotnet/pull/1471#discussion_r529422644
Good catch.
(JFYI it's part of test applications, not SDK, so I'd not block the release because of it)

MatkovIvan (approved):

Apps/Contoso.Android.Puppet/ModulePages/AppCenterFragment.cs
MatkovIvan: It will be addressed separately, I approved this PR

EatonZ:
Hello, 2 things:
1. When will 4.1.0 be published to NuGet?
2. Will passing 0 to `SetMaxStorageSizeAsync` essentially disable creation/usage of sqlite databases?

MatkovIvan:
Hi @EatonZ,
1. It had some issues, we'll publish it after holydays
2. No. It has the same behaviour [like other platforms](https://docs.microsoft.com/en-us/appcenter/sdk/other-apis/ios#storage-size):
   > If you don't set the max storage size, the SDK uses a default max size of 10 MB. The minimum size you're allowed to set is 20 KB.

Issue #1627, opened 2022-02-17
[Windows] The initialization of App Center underlying services fails, due to a deadlock.


dahlita:
### **Description**

This issue seems to manifest when starting more than one services, e.g `Analytics` then `Crashes`.
The 1st service starts successfully, but the 2nd one fails due to a deadlock.
When the deadlock happens while `AppCenter.Start(params Type[] services)` was invoked on the UI thread, the entire app UI freezes.
Moving the initialization on a background thread, helps keep the UI responsive, but that thread then gets frozen, and blocks all other threads executing events from the 1st service that started successfully.

### **Repro Steps**

Simply invoke
  `AppCenter.Start("<app secret here>", typeof(Analytics), typeof(Crashes))`;

Or start them one after the other with:
```
 AppCenter.Configure("<app secret here>") 
 AppCenter.Start(typeof(Analytics));
 AppCenter.Start(typeof(Crashes));
```

### **Platform details**

1. What is your app platform?
    - Windows desktop
2. Which SDK version are you using?
    - 4.5.0
3. Which OS version did you experience the issue on?
    - Windows 10.0.19043
4. What third party libraries are you using?
    - I can reproduce the issue in a small sample app that just references
     `Microsoft.AppCenter.Analytics` and `Microsoft.AppCenter.Crashes`

### **More details on the deadlock with the call stack**

While instantiating the service `Channel`, the thread acquires a lock via `_mutex.GetLock()`, then queues a BG task to get the logs count
(the result of this task will later be used in another continuation task to update the initial value of `_pendingLogCount` and release the above lock).
Then the thread continues to check for pending logs, via `Channel.CheckPendingLogsInternal(State state)`, attempts to acquire the lock again and waits for the previous one to be released before continuing its execution.
Meanwhile, the 1st service that started successfully is executing its event logs  on another BG thread, but is blocked because it needs the global lock `AppCenterLock` held by the thread instantiating the 2nd service.
At the same time, the Task scheduler is waiting for the ongoing tasks to complete, and related threads to be released, so that it can execute other scheduled ones.
And one of the tasks that is still queued - still in the `Scheduled` status - is the task responsible for populating `_pendingLogCount` and release the lock above, acquired in the `Channel.ctor(...)`.

Below are screenshots showing the process dump analysis of blocked threads / tasks as well as scheduled ones, when the issue occurs:
 
[...]

PR #1628, opened 2022-02-18
Fix a deadlock that occurs during the services initialization

dahliaingabire-okta:

## Description

Issue:
Currently, when the task responsible for releasing the lock acquired during the `Channel` instantiation is not quickly executed and remains in the scheduler queue while there are other tasks executing concurrently, it can lead to deadlocks.
More details in https://github.com/microsoft/appcenter-sdk-dotnet/issues/1627

Suggestion:
Make the task that initializes  `_pendingCountLog` available to private methods  so that it can be `awaited` upon when needing to evaluate and / or update the field.

## Related PRs or issues

Fixes the issue reported in https://github.com/microsoft/appcenter-sdk-dotnet/issues/1627

## Misc

Add what's missing, notes on what you tested, additional thoughts or questions.


SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Channel/Channel.cs
```diff
@@ -635,11 +648,11 @@ public void SetNetworkRequestAllowed(bool isAllowed)
         {
             if (isAllowed)
             {
-                Resume(_mutex.State, false);
+                Resume(_mutex.State, false).ConfigureAwait(false);
```
DmitriyKirakosyan, line 651: @dahliaingabire-okta , I think we don't need `.ConfigureAwait(false)` if we don't `await` it. Same for the two places below.

dahliaingabire-okta: Right! Removed; thank you! Will you review again?

DmitriyKirakosyan (approved):

SDK/AppCenter/Microsoft.AppCenter.Windows.Shared/Channel/Channel.cs
DmitriyKirakosyan: Sorry for the delay. Approved. There are some checks failed, I'll take a look and get back to you.

DmitriyKirakosyan:
@dahliaingabire-okta , thank you for your contribution!

PR #1642, opened 2022-04-26
Improve/fix CreateModelExceptionAndBinaries

EatonZ:
There appears to be a small bug in the `CreateModelExceptionAndBinaries` method. If it processes an `AggregateException`, it will actually add the inner exception twice because of the second if block. I confirmed it by debugging - at the end, `modelException.InnerExceptions` has 2 list items that are the same. Since the first if block processes the inner exceptions of the `AggregateException`, it's not necessary to execute the second if block unless the exception is **not** an `AggregateException`.

If this was by design for some reason, let me know. Seems like a bug, though. I don't see why you would want to duplicate the same inner exception.

### DIFF
--- a/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.WindowsDesktop/Utils/ErrorLogHelperPart.cs
+++ b/SDK/AppCenterCrashes/Microsoft.AppCenter.Crashes.WindowsDesktop/Utils/ErrorLogHelperPart.cs
@@ -29,7 +29,7 @@ namespace Microsoft.AppCenter.Crashes.Utils
                     }
                 }
             }
-            if (exception.InnerException != null)
+            else if (exception.InnerException != null)
             {
                 modelException.InnerExceptions = modelException.InnerExceptions ?? new List<ModelException>();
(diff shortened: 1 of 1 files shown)


AnatolyPristensky:
Hello @EatonZ and thank you for your contribution! That's looks like a bug. 
Could you please add a unit test that indicates an issue?

AnatolyPristensky:
Hello @EatonZ and thank you for your contribution! That's looks like a bug. 
Could you please add a unit test that indicates an issue?

EatonZ:
@AnatolyPristensky Ok, so this was indeed unintended behavior? How was the server API processing the duplicate exception?

As for a test, I'll try and take a look when I have time..

DmitriyKirakosyan:
Hi @EatonZ,

> Ok, so this was indeed unintended behavior? How was the server API processing the duplicate exception?

Yes, this was a bug, thank you for catching it and preparing the fix! The backend doesn't have any logic for a distinction of such duplicates. It will end up with duplicated inner exceptions on portal, if you look the exception's Raw.

Please let us know if you need any help with unit tests.

AnatolyPristensky (approved):

MatkovIvan (approved):

Issue #1652, opened 2022-06-15
WPF DeviceInformationHelper doesn't handle unsupported platforms properly


FrayxRulez:
### **Description**

WPF `DeviceInformationHelper` is using `System.Management` NuGet package to load info about the device, but this seems to fail on some cases because `System.Management` hooks some built-in .NET Framework native DLL (`wminet_utils.dll`) that might be outdated in some cases.
If the library fails to load, `System.Management` throws a `PlatformNotSupportedException` that's not actually being catched by `DeviceInformationHelper` causing the library to never initialize and to rethrow the exception over and over again.

### **Repro Steps**

The easier way to reproduce this is to test the library on some ancient Windows 10 installation, we tried on version `10240` to be specific.
Invoking `AppCenter.Start("xxx", typeof(Crashes));` seems to be enough to trigger the exception.

### **Details**

We're on WPF, .NET 6, and SDK version 4.5.1.
Windows 10, version 10240 (the issue seem to exist on Windows 7 and 8 as well).

DmitriyKirakosyan:
Hi @FrayxRulez , thank you for reporting this issue, it looks like a bug indeed. Could you clarify this statement:

> If the library fails to load, System.Management throws a PlatformNotSupportedException that's not actually being catched by DeviceInformationHelper causing the library to never initialize and to rethrow the exception over and over again.

Does it mean that AppCenter SDK actually crashes your app, or it just fails to initialize and as a result it is not working on the older windows builds?

DmitriyKirakosyan:
The fix has been released in v4.5.3. Closing.

PR #1653, opened 2022-06-15
Fix missing catch for unsupported platform

FrayxRulez:

## Description

Just adding a catch for unsupported platform exceptions on Windows Desktop.

## Related PRs or issues

This is connected to #1652

### DIFF
--- a/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
+++ b/SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
@@ -81,6 +81,11 @@ namespace Microsoft.AppCenter.Utils
                 AppCenterLog.Warn(AppCenterLog.LogTag, "Failed to get device model. Make sure that WMI service is enabled.", exception);
                 return string.Empty;
             }
+            catch (PlatformNotSupportedException exception)
+            {
+                AppCenterLog.Warn(AppCenterLog.LogTag, "Failed to get device model. Make sure that .NET Framework is up to date.", exception);
+                return string.Empty;
+            }
             return string.Empty;
         }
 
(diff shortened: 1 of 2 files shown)


DmitriyKirakosyan:
Hi @FrayxRulez , thank you for your contribution!
Could you also update the Changelog file?

SDK/AppCenter/Microsoft.AppCenter.WindowsDesktop/Utils/DeviceInformationHelper.cs
```diff
@@ -81,6 +81,11 @@ protected override string GetDeviceModel()
                 AppCenterLog.Warn(AppCenterLog.LogTag, "Failed to get device model. Make sure that WMI service is enabled.", exception);
                 return string.Empty;
             }
+            catch (PlatformNotSupportedException exception)
+            {
+                AppCenterLog.Warn(AppCenterLog.LogTag, "Failed to get device model. Make sure that .NET Framework is up to date.", exception);
```
DmitriyKirakosyan, line 86: This message can be confusing. Does the exception occur on a specific versions of .NET Framework? We need to specify the minimum supported version in that case.
Same question for the similar logs below.

FrayxRulez: I'm afraid I can't answer this question, I fiddled a bit with ILSpy in the .NET DLL throwing the exception and it seems to just throw if any of the needed symbols are not found in the Win32 DLL, without providing any further information...

FrayxRulez: So the idea is that even the `System.Management` package coming from NuGet tries to load `C:\Windows\Microsoft.NET\Framework64\v4.0.30319\WMINet_Utils.dll` and this DLL comes with Windows.

DmitriyKirakosyan: Got it, thanks.

DmitriyKirakosyan (approved):

FrayxRulez:
@DmitriyKirakosyan actually not sure about what to do with the failing tests, I highly doubt they're failing because of my changes but I have no way to check, as I have no access to the DevOps logs.

aleksandr-dorofeev (approved):

PR #1705, opened 2022-11-24
Register AppDelegate dependency.

ternovtsi:
## Description
IClearCrashClick was not registered in DependencyService causing NRE when trying to call 'ClearCrashButton'

### DIFF
--- a/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.MacOS/AppDelegate.cs
+++ b/Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.MacOS/AppDelegate.cs
@@ -4,7 +4,10 @@
 using AppKit;
 using Foundation;
 using Xamarin.Forms.Platform.MacOS;
+using Xamarin.Forms;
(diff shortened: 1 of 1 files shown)


Apps/Contoso.Forms.Demo/Contoso.Forms.Demo.MacOS/AppDelegate.cs
```diff
@@ -4,7 +4,10 @@
 using AppKit;
 using Foundation;
 using Xamarin.Forms.Platform.MacOS;
+using Xamarin.Forms;
```
DmitriyKirakosyan, line 7: We do we need these usings?

ternovtsi: using Xamarin.Forms for 'Dependency'
using Contoso.Forms.Demo.MacOS; for 'AppDelegate'

DmitriyKirakosyan (approved):

PR #1800, opened 2024-12-17
Remove old Xamarin demo apps

DordeDimitrijev:
## Description
Removed old demo Xamarin apps and it's references in files and scripts. Updated MAUI demo config so it can be buildable and debuggable with local appcenter-dotnet sdk.

## Related PRs or issues

[#AB107104
](https://dev.azure.com/msmobilecenter/Mobile-Center/_boards/board/t/Mobile%20SDK%20team/Backlog%20Items?text=dot&workitem=107104)


scripts/update-app-secrets.sh
```diff
@@ -1,52 +0,0 @@
-#!/bin/bash
```
DmitriyKirakosyan, line 1: Why did you remove the file? Isn't (shouldn't) it used for MAUI demo?

DordeDimitrijev: Yes, it MAUI is also mentioned in the script but I'm not sure was even used for MAUI. I couldn't find any usage of the script for the MAUI demo app. I could leave the script and remove all xamarin related stuff?

DmitriyKirakosyan: It should update secrets. See https://github.com/microsoft/appcenter-sdk-dotnet/blob/56f71941990906039841d50257cdad302e7f169b/Apps/Contoso.MAUI.Demo/App.xaml.cs#L118

DordeDimitrijev: I've reverted it and refactored it a bit since we've removed xamarin apps

DmitriyKirakosyan (approved):

